PR #9, opened 2015-07-24
oauth2,oidc: add client.RefreshToken() method.

yifan-gu:
client.RefreshToken() takes a refresh token and request a new token
from the Token Endpoint.

Also added 'refresh_token' and 'scope' field in oauth2.TokenResult.

/cc @bobbyrullo

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -19,9 +19,10 @@ const (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


philips:
lgtm, @jonboulle is final reviewer.

yifan-gu:
@bobbyrullo Documented :)

PR #10, opened 2015-07-29
token extractor utils and url validation

sym3tri:
This obviates some of the token extraction boilerplate when using this library.
Also adds stricter URL validation which was causing uncaught errors to bubble up during actual usage

### DIFF
--- a/oidc/client.go
+++ b/oidc/client.go
@@ -66,11 +66,16 @@ type ClientConfig struct {
 }
 
 func NewClient(cfg ClientConfig) (*Client, error) {
+	ru, err := phttp.ParseNonEmptyURL(cfg.RedirectURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid redirect URL: %v", err)
+	}
+
 	c := Client{
 		credentials:    cfg.Credentials,
 		httpClient:     cfg.HTTPClient,
 		scope:          cfg.Scope,
-		redirectURL:    cfg.RedirectURL,
+		redirectURL:    ru.String(),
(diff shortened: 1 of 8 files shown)


bobbyrullo:
minor things, otherwise LGTM

sym3tri:
addressed comments and rebased

oidc/client.go
```diff
@@ -66,11 +66,16 @@ type ClientConfig struct {
 }
 
 func NewClient(cfg ClientConfig) (*Client, error) {
+	ru, err := phttp.ParseNonEmptyURL(cfg.RedirectURL)
```
yifan-gu: @sym3tri Later for the party, but why redirectURL is required here? E.g. I think I can use the client for just token verification or token refreshing, which doesn't need the redirectURL info.

sym3tri: Clients need a callback URL in order to get tokens from the OIDC server. That's the only way a token can be obtained. Why would a client ever need to validate a token without allowing it to be generated?

Are there any actual use-cases requiring verification without generation? Even if a user is pasting in a token + refresh token, I think the client still needs to support logging in via the browser and the 3-legged OAuth2 flow if available.

yifan-gu: @sym3tri My use case here is the kube-apiserver acts as a client, but it doesn't handle login or cause to generate any tokens, it just verifies the token.

sym3tri: I see... that makes sense. Feel free to remove this validation if it helps your case.

yifan-gu: Good to know, thanks :)

PR #19, opened 2015-10-19
jose/oidc: allow string array in claims['aud'].

### LINKED ISSUE #18: VerifyClaims should tread 'aud' as an array of string
Currently it treats it as a single string. and the error message is unclear (what's the expected valued and actual value?)

/cc @polvi

yifan-gu:
Fix #18 
cc @bobbyrullo @polvi

### DIFF
--- a/jose/claims.go
+++ b/jose/claims.go
@@ -26,6 +26,32 @@ func (c Claims) StringClaim(name string) (string, bool, error) {
 	return v, true, nil
 }
 
+func (c Claims) StringsClaim(name string) ([]string, bool, error) {
+	cl, ok := c[name]
+	if !ok {
+		return nil, false, nil
+	}
+
+	if v, ok := cl.([]string); ok {
+		return v, true, nil
+	}
+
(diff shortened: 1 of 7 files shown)


yifan-gu:
As in the spec, the 'aud' should allow an string array: http://openid.net/specs/openid-connect-core-1_0.html#IDToken

polvi:
nice!

yifan-gu:
@bobbyrullo naming fixed

yifan-gu:
Next step is:
- pass the `aud` to server, this should be implemented in go-oidc
- pass the `aud` back in the id token claims, this should be implemented in dex

For the first one I didn't find any specification on how to pass it, so I assume it's up to us.
I plan to add a field in the client config and oauthclient config, called `audience`. Then embed the field in the url, similar to what we do for [scope](https://github.com/coreos/go-oidc/blob/master/oauth2/oauth2.go#L134) 

@bobbyrullo

bobbyrullo:
LGTM

PR #20, opened 2015-10-21
jose: Use claims type instead of map for NewSignedJWT

ecnahc515:
Use the claims type rather than passing a map to `NewSignedJWT(...)`

### DIFF
--- a/jose/jwt.go
+++ b/jose/jwt.go
@@ -63,13 +63,13 @@ func (j *JWT) Encode() string {
 	return strings.Join([]string{d, s}, ".")
 }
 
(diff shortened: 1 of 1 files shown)


ecnahc515:
cc @bobbyrullo just a ergonomics fix, it make's the JWT constructors more consistent.

ecnahc515:
I also just noticed, in `NewJWT()` we return a `JWT`, but in `NewSignedJWT` we return a `*JWT`, and in `ParseJWT()` we return a `JWT`. What's the preferred approach here? It seems since all the methods act on pointer receivers we should be returning `*JWT` everywhere.

bobbyrullo:
LGTM

PR #22, opened 2015-11-02
key: impl json.Marshaler/Unmarshaler for PublicKey

ashishgandhi:


bobbyrullo:
Good point. LGTM

bobbyrullo:
Thanks @ashishgandhi !

PR #24, opened 2015-12-03
oauth2: don't fail if the token response does not expire

ericchiang:
Some oauth2 implementations don't have an 'expires' or 'expires_in'
field, specifically GitHub. Don't fail if this happens.

Needed to use this package for coreos/dex#191

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -250,9 +250,11 @@ func parseTokenResponse(resp *http.Response) (result TokenResponse, err error) {
 		if e == "" {
 			e = vals.Get("expires")
 		}
-		result.Expires, err = strconv.Atoi(e)
-		if err != nil {
-			return
+		if e != "" {
+			result.Expires, err = strconv.Atoi(e)
+			if err != nil {
+				return
+			}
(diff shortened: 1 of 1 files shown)


joeatwork:
This looks good to me, but we should probably ask somebody a bit more security savvy ( @gtank ? wanna take a look?)

(Security-savvy reviewer take note - this change is to work with a real-world provider, and echoes the behavior of https://github.com/golang/oauth2 - the relevant bit of the spec, if I google correctly, is here 
https://tools.ietf.org/html/rfc6749#section-4.2.2 )

ericchiang:
Additional notes. We ever use the `Expired` field in this package or the provided @joeatwork mentioned. Further more, in the OIDC spec 'expires_in' is marked optional in the OIDC Core Spec [Section 3.2.2.5](http://openid.net/specs/openid-connect-core-1_0.html#rfc.section.3.2.2.5)

gtank:
This looks reasonable. It's clearly OPTIONAL in both specs and OAuth2.0 explicitly says that this is the server's responsibility, not the client library's.

Issue #25, opened 2015-12-04
oauth2: errors not reported during token request 


ericchiang:
Currently using this package to implement github and bitbucket oauth2 clients. Noticed that some server response errors are either being misreported or lacking detail during the token exchange.

Full request dump [here](https://gist.github.com/ericchiang/6e82e497ec45ea2aedbf)

When using an incorrect client secret:

Bitbucket:
* Returns an `error_description` field which is optional in the [oauth2 spec](https://tools.ietf.org/html/rfc6749#section-4.1.2.1) but this package doesn't support.

GitHub
* Returns a 200 status when it serves an error so the error is [never detected](https://github.com/coreos/go-oidc/blob/562ae81124e0ae327b444d609a597a35f3521027/oauth2/oauth2.go#L224-L227) by this package.
* Returns an `error_description` field (same problem as bitbucket).
* Returns errors x-www-form-urlencoded not JSON encoded so would result in an [unrecognized error](https://github.com/coreos/go-oidc/blob/562ae81124e0ae327b444d609a597a35f3521027/oauth2/error.go#L36).

PR #26, opened 2015-12-05
oauth2: adjusting error detection and reporting

ericchiang:
Based on experiencince with github and bitbucket oauth (see #25), change error detection to catch the following cases on token responses.

* Add `error_description` field to oauth2.Error (bitbucket & github).
* Catch errors with 2XX HTTP statuses (github).
* Handle x-www-form-urlencoded errors (github).

Additional fixes

* Fixes `expires_in` parsing for JSON responses.

Bug in original code:

```
b := make(map[string]interface{})
data := []byte(`{"expires_in":1000}`)
if err := json.Unmarshal(data, &b); err != nil {
	panic(err)
}
if val, ok := b["expires_in"].(int); ok {
	// you'll never get here
	fmt.Println("expires in", val)
}
	
fmt.Printf("%T\n", b["expires_in"]) // float64
```


bobbyrullo:
LGTM

Issue #17, opened 2015-10-09
Race seen in client sync


liggitt:
reported in https://github.com/openshift/origin/issues/5025#issuecomment-146807693
```
==================

WARNING: DATA RACE

Read by goroutine 53:

  github.com/coreos/go-oidc/oidc.(*Client).VerifyJWT()

      /home/travis/gopath/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/coreos/go-oidc/oidc/client.go:285 +0x261

  github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/plugin/pkg/auth/authenticator/token/oidc.(*OIDCAuthenticator).AuthenticateToken()

      github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/plugin/pkg/auth/authenticator/token/oidc/_test/_obj_test/oidc.go:140 +0x1bb

  github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/plugin/pkg/auth/authenticator/token/oidc.TestOIDCAuthentication()

      /home/travis/gopath/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/Godeps/_workspace/src/k8s.io/kubernetes/plugin/pkg/auth/authenticator/token/oidc/oidc_test.go:378 +0x1654

  testing.tRunner()

      /tmp/workdir/go/src/testing/testing.go:456 +0xdc

Previous write by goroutine 90:

  github.com/coreos/go-oidc/oidc.(*providerConfigRepo).Set()

      /home/travis/gopath/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/coreos/go-oidc/oidc/client.go:194 +0x86

  github.com/coreos/go-oidc/oidc.(*ProviderConfigSyncer).sync()

      /home/travis/gopath/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/coreos/go-oidc/oidc/provider.go:111 +0x15a

  github.com/coreos/go-oidc/oidc.(*ProviderConfigSyncer).(github.com/coreos/go-oidc/oidc.sync)-fm()

      /home/travis/gopath/src/github.com/openshift/origin/Godeps/_workspace/src/github.com/coreos/go-oidc/oidc/provider.go:95 +0x3b

  github.com/coreos/go-oidc/oidc.(*pcsStepNext).step()

[...]

PR #27, opened 2015-12-08
oidc: guard client provider config against race conditions

ericchiang:
Add a RWMutex around a client's provider config to prevent race conditions when updating after syncing.

Test added which triggered the race condition before this change.

cc @yifan-gu, @bobbyrullo

### DIFF
--- a/oidc/client_race_test.go
+++ b/oidc/client_race_test.go
@@ -1,0 +1,70 @@
+// This file contains tests which depend on the race detector being enabled.
+// +build race
+
+package oidc
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+type testProvider struct {
+	baseURL string
+}
+
+func (p *testProvider) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


ericchiang:
@bobbyrullo this has been updated with the discussed changes.

oidc/client_race_test.go
```diff
@@ -0,0 +1,61 @@
...
+	// SyncProviderConfig beings a goroutine which writes to the client's provider config.
+	c := cli.SyncProviderConfig(s.URL)
+	defer func() {
+		// stop the background process
+		c <- struct{}{}
+	}()
+
+	// Iterate serveral times to increase the likelyhood of triggering the race detector.
+	for i := 0; i < 100; i++ {
+		time.Sleep(5 * time.Millisecond)
```
yifan-gu: The next sync will happen in `MinimumProviderConfigSyncInterval` which is 1 min. maybe we make that tunable for test ?

yifan-gu:
LGTM overall! Thank you @ericchiang :)

yifan-gu:
This should fix #17

oidc/client_race_test.go
ericchiang: done :+1:

PR #31, opened 2015-12-09
key: increase default RSA key size from 1024 to 2048 bits

### LINKED ISSUE #15: go-oidc only generates 1024-bit RSA keys
This key size is [hardcoded](https://github.com/coreos/go-oidc/blob/master/key/key.go#L124 ). While 1024-bit keys aren't currently broken, they are notably weaker than we should choose for a new system.

I suggest that for now we bump this to 2048, and in the future consider moving to elliptic curve signatures.

ericchiang:
Per @gtank's recommendations, increase the default RSA key size in
GeneratePrivateKey from 1024 to 2048 bits.

Closes #15

### DIFF
--- a/key/key.go
+++ b/key/key.go
@@ -135,7 +135,7 @@ func (s *PrivateKeySet) Active() *PrivateKey {
 type GeneratePrivateKeyFunc func() (*PrivateKey, error)
 
 func GeneratePrivateKey() (*PrivateKey, error) {
-	pk, err := rsa.GenerateKey(rand.Reader, 1024)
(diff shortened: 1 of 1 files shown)


gtank:
The reason I didn't do this already is because I haven't thought through the migration implications (if there are any). If that's all good (@bobbyrullo?) then LGTM.

joeatwork:
My (not very deep) thoughts about this - since this change only applies to new keys (and clients of those keys really shouldn't be making assumptions about their size) this change LGTM. I'd still love if @bobbyrullo weighed in, though.

bobbyrullo:
LGTM - sorry for taking so long to look at this

PR #35, opened 2015-12-17
oidc: Add WaitUntilProviderSync to client

bobbyrullo:
This way we can set an initial providerconfig before we start syncing.

### DIFF
--- a/oidc/client.go
+++ b/oidc/client.go
@@ -101,32 +101,10 @@ type Client struct {
 	redirectURL    string
 	scope          []string
 	keySet         key.PublicKeySet
+	providerSyncer *ProviderConfigSyncer
 
 	keySetSyncMutex sync.RWMutex
 	lastKeySetSync  time.Time
-}
-
-type providerConfigRepo struct {
-	mu     sync.RWMutex
-	config ProviderConfig // do not access directly, use Get()
-}
-
-func newProviderConfigRepo(pc ProviderConfig) *providerConfigRepo {
-	return &providerConfigRepo{sync.RWMutex{}, pc}
-}
-
-// returns an error to implement ProviderConfigSetter
-func (r *providerConfigRepo) Set(cfg ProviderConfig) error {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-	r.config = cfg
-	return nil
-}
-
-func (r *providerConfigRepo) Get() ProviderConfig {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
-	return r.config
 }
 
 func (c *Client) Healthy() error {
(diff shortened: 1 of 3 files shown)


ericchiang:
Since we're bringing this up. Are there any scenarios where we'd want to create a client but not have it fetch a provider config (assuming you don't give it one to start with)? Can we just have `NewClient` fetch the provider config?

I'm asking because it seems like there are a lot of bugs stemming from the fact that `NewClient` doesn't do this.

bobbyrullo:
That's a good point. I can't think of any per se, but it does make testing a bit harder - see the auth_test in bridge - it creates a NewClient() and then tests some of the redirect methods. But the discovery url it's given is fake (auth.example.com) so the sync can't work. If we block in NewClient(), then it just hangs forever.

Another alternative is to make SyncProviderConfig() block until the initial sync is performed. But that might annoy existing code which doesn't expect it to block.

Another Another alternative is to add a "WaitForInitialConfig" to the client, which blocks until there's a valid one. That gives users the choice. What do you think?

ericchiang:
@bobbyrullo yeah that sound like a good plan.

ericchiang:
My issue is that

1) WaitForInitialConfig still doesn't return an error.
2) With additional methods on the client it becomes harder to use this package correctly.

I think what the problem we're actually trying to solve is _"I have the issuerURL but not the provider config. When I create a new client I expect it to populate the config and tell me if my issuerURL is correct."_

This fix does work, but could we go with something more direct so NewClient() blocks?

```diff
index 7633023..c4d2c7b 100644
--- a/oidc/client.go
+++ b/oidc/client.go
@@ -63,6 +63,7 @@ type ClientConfig struct {
        RedirectURL    string
        ProviderConfig ProviderConfig
        KeySet         key.PublicKeySet
+       IssuerURL      string
 }
 
 func NewClient(cfg ClientConfig) (*Client, error) {
@@ -73,6 +74,16 @@ func NewClient(cfg ClientConfig) (*Client, error) {
                return nil, fmt.Errorf("invalid redirect URL: %v", err)
        }
 
+       if cfg.ProviderConfig.Empty() {
+               if cfg.IssuerURL == "" {
+                       return nil, errors.New("must provide an issuer URL or an existing provider config")
+               }
+               cfg.ProviderConfig, err = FetchProviderConfig(cfg.HTTPClient, cfg.IssuerURL)
+               if err != nil {
+                       return nil, err
+               }
+       }
+
        c := Client{
                credentials:    cfg.Credentials,
                httpClient:     cfg.HTTPClient,

```

bobbyrullo:
> 1) WaitForInitialConfig still doesn't return an error.
It doesn't return an error...because it cannot. It keeps retrying.

> 2) With additional methods on the client it becomes harder to use this package correctly.
Marginally harder. You're not *required* to use the WaitUntilIntialSync method.

> This fix does work, but could we go with something more direct so NewClient() blocks?
I don't want to put blocking behavior in  `New*` functions because of testability issues.

How about instead we make the SyncProviderConfig block until the first sync?

ericchiang:
LGTM

PR #38, opened 2015-12-30
oidc: update client metadata and provider config to support dynamic registration spec

ericchiang:
Update this package to hold structs required for dynamic client registration. ClientMetadata and ClientRegistrationResponse.

This commit is breaking. It updates ClientMetadata to be JSON compatible with the spec by
  1. renaming `RedirectURLs` to `RedirectURIs` as per [the spec](http://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata)
  2. changing that field's type from `[]url.URL` to `[]string` for JSON serialization 

The tests affected by this change have been rectified. We will have to make a few changes to dex when updating this package (which we'll have to anyway to store the added fields in the client metadata).

@bobbyrullo I don't know if this will fly, but I'd like to get your feedback nonetheless. Would it be better to keep `RedirectURLs` and list it deprecated?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ go:
   - 1.5.2
 
 install:
- - go get -v ./...
+ - go get -v -t ./...
  - go get golang.org/x/tools/cmd/cover
  - go get golang.org/x/tools/cmd/vet
 
--- a/jose/jose.go
+++ b/jose/jose.go
@@ -13,6 +13,57 @@ const (
 	HeaderKeyID        = "kid"
 )
 
+const (
+	// Encryption Algorithm Header Parameter Values for JWS
+	// See: https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40#page-6
+	AlgHS256 = "HS256"
+	AlgHS384 = "HS384"
+	AlgHS512 = "HS512"
+	AlgRS256 = "RS256"
+	AlgRS384 = "RS384"
+	AlgRS512 = "RS512"
+	AlgES256 = "ES256"
+	AlgES384 = "ES384"
+	AlgES512 = "ES512"
+	AlgPS256 = "PS256"
+	AlgPS384 = "PS384"
+	AlgPS512 = "PS512"
+	AlgNone  = "none"
+)
+
+const (
+	// Algorithm Header Parameter Values for JWE
+	// See: https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40#section-4.1
+	AlgRSA15            = "RSA1_5"
+	AlgRSAOAEP          = "RSA-OAEP"
+	AlgRSAOAEP256       = "RSA-OAEP-256"
+	AlgA128KW           = "A128KW"
+	AlgA192KW           = "A192KW"
+	AlgA256KW           = "A256KW"
+	AlgDir              = "dir"
+	AlgECDHES           = "ECDH-ES"
+	AlgECDHESA128KW     = "ECDH-ES+A128KW"
+	AlgECDHESA192KW     = "ECDH-ES+A192KW"
+	AlgECDHESA256KW     = "ECDH-ES+A256KW"
+	AlgA128GCMKW        = "A128GCMKW"
+	AlgA192GCMKW        = "A192GCMKW"
+	AlgA256GCMKW        = "A256GCMKW"
+	AlgPBES2HS256A128KW = "PBES2-HS256+A128KW"
+	AlgPBES2HS384A192KW = "PBES2-HS384+A192KW"
+	AlgPBES2HS512A256KW = "PBES2-HS512+A256KW"
+)
+
+const (
+	// Encryption Algorithm Header Parameter Values for JWE
+	// See: https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40#page-22
+	EncA128CBCHS256 = "A128CBC-HS256"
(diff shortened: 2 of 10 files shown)


ericchiang:
Additionally considerations: Since we store client metadata as text in dex, this change should be negligible on the database side.

Storing it [here](https://github.com/coreos/dex/blob/7c8e25f2923e71b9736c14abdf25c9acb69c9984/db/client.go#L46-L64) becomes more trivial.

```diff
diff --git a/db/client.go b/db/client.go
index 89a1667..b5c7f22 100644
--- a/db/client.go
+++ b/db/client.go
@@ -49,7 +49,7 @@ func newClientIdentityModel(id string, secret []byte, meta *oidc.ClientMetadata)
                return nil, err
        }
 
-       bmeta, err := json.Marshal(newClientMetadataJSON(meta))
+       bmeta, err := json.Marshal(meta)
        if err != nil {
                return nil, err
        }
``` 

Maintaining backwards compatibility in unmarshaling [here](https://github.com/coreos/dex/blob/7c8e25f2923e71b9736c14abdf25c9acb69c9984/db/client.go#L85-L103) is also fairly easy (though might pose issues for rolling updates).

```diff
$ git diff
diff --git a/db/client.go b/db/client.go
index 89a1667..33b7df9 100644
--- a/db/client.go
+++ b/db/client.go
@@ -84,22 +83,14 @@ func newClientMetadataJSON(cm *oidc.ClientMetadata) *clientMetadataJSON {
 
 type clientMetadataJSON struct {
        RedirectURLs []string `json:"redirectURLs"`
+       *oidc.ClientMetadata
 }
 
 func (cmj clientMetadataJSON) ClientMetadata() (*oidc.ClientMetadata, error) {
-       cm := oidc.ClientMetadata{
-               RedirectURLs: make([]url.URL, len(cmj.RedirectURLs)),
+       if len(cmj.RedirectURLs) > 0 {
+               cmd.ClientMetadata.RedirectURIs = cmj.RedirectURLs
        }
-
-       for i, us := range cmj.RedirectURLs {
-               up, err := url.Parse(us)
-               if err != nil {
-                       return nil, err
-               }
-               cm.RedirectURLs[i] = *up
-       }
-
-       return &cm, nil
+       return cmd.ClientMetadata.RedirectURIs, nil
 }
```

ericchiang:
@bobbyrullo as a general comment on `url.URL` vs `string`. For the dynamic registration stuff, I'm going to be marshaling and unmarshaling these structs for client requests and server responses. `url.URL` doesn't encode nicely when using `encoding/json`.

So we can either:
* define custom `MarshalJSON()` and `UnmarshalJSON()` for these structs or
* use `string` rather than `url.URL` and let `Validate()` ensure they're URLs

I choose the later but would be happy to do the former.

ericchiang:
This has been updated to give a lot more documentation and verification around fields. I've tried to address some additional concerns in the various comment threads.

I need to add a bunch of tests, but I'd like to get some feedback before I dive in :)

[link to outdated diff with discussion](https://github.com/ericchiang/go-oidc/commit/72354f257be133ebb2f871a36d4d46a313a5f721)

ericchiang:
@bobbyrullo this is ready for another review.

ericchiang:
Additional note: [ProviderConfig](https://github.com/coreos/go-oidc/blob/6c7d98c19949818b4d10b076f70a6f278552fa84/oidc/provider.go#L32-L43) needs to be updated as well to include `registration_endpoint`. Do we want to convert the string fields to `*url.URL`s like we've done for ClientMetadata? It's a little more integrated with the package, so package uses might feel that breaking change a lot more.

bobbyrullo:
Looks good, just some cosmetic stuff. Question: will this break what's currently stored within the database? I'm guessing/hoping it wont because it's all additive right?

ericchiang:
Field names updated, and comments added.

> will this break what's currently stored within the database? I'm guessing/hoping it wont because it's all additive right?

We corrected `redirect_urls` to `redirect_uris` (what OIDC actually uses), so it is breaking. When reading values from the DB, we'll have to check `redirect_urls` as well as `redirect_uris` for values.

bobbyrullo:
uggh. we are using default JSON encoding in dex. we'll need to have a shim in there so that it recognizes  redirectURLs and redirect_uris

bobbyrullo:
ok but be careful in dex - the old value is not "redirect-urls", it's "redirectUrls" since we were just using default JSON names.

bobbyrullo:
LGTM

ericchiang:
@bobbyrullo This has been updated.

Biggest changes are `ProviderConfig` has been updated to match `ClientMetadata`. Both have `Valid` methods which check OIDC compliance and `Defaults` methods which fill omitted fields with defaults values.

This work is largely to support a [`(ProviderConfig).Supports(ClientMetadata)`](https://github.com/coreos/go-oidc/blob/0db2eca31dbf51af2bdaaef3821d32874af8a7a3/oidc/provider.go#L390) method which indicates if a provider supports the options a client request, which I saw as an important


```go
type Server struct {
    ...
    cfg oidc.ProviderConfig
}

func (s *Server) handleRegistration(w http.ResponseWriter, r *http.Request) {
    ...

    var clientMetadata oidc.ClientMetadata
    if err := json.NewDecoder.Decode(&clientMetadata); err != nil {
        // failed to unmarshal client metadata requests
    }

    if err := s.cfg.Supports(clientMetadata): err != nil {
        // server does not support some of the options the client requested
    }

    ...

}

```

PR #47, opened 2016-02-12
provider: Use path.Join() to construct the URL.

yifan-gu:
Should fix https://github.com/kubernetes/kubernetes/issues/20476

cc @bobbyrullo @ericchiang


yifan-gu:
close in favor of https://github.com/kubernetes/kubernetes/pull/21128

PR #49, opened 2016-02-20
add revocation_endpoint in provider config

### LINKED ISSUE #48: add revocation_endpoint in provider config

yutongp:
I know this is not define in spec, but usually there is a revocation endpoint.
https://login.salesforce.com/.well-known/openid-configuration
https://accounts.google.com/.well-known/openid-configuration

Fixes #48

### DIFF
--- a/oidc/provider.go
+++ b/oidc/provider.go
@@ -68,6 +68,7 @@ type ProviderConfig struct {
 	UserInfoEndpoint     *url.URL
 	KeysEndpoint         *url.URL // Required
 	RegistrationEndpoint *url.URL
+	RevocationEndpoint   *url.URL
 
 	// Servers MAY choose not to advertise some supported scope values even when this
 	// parameter is used, although those defined in OpenID Core SHOULD be listed, if supported.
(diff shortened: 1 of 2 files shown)


ericchiang:
Would it better to allow the provider metadata to decode/encode additional parameters?

```go
type ProviderConfig struct {
    ...
    Extensions map[string]interface{} // Additional fields parsed from the ProviderMetadata
}
```

These are needed by other uses of provider metadata (e.g. [here](https://openid.net/specs/openid-connect-session-1_0.html)) and I don't want to get in the habit of adding struct fields that aren't defined in the spec.

yutongp:
Fair. I think that will serve the purpose well.

yutongp:
wait. How would you marshal it to
```
{
  ...
  "revocation_endpoint": "https://server.example.com/connect/revoke"
}
```
instead of 
```
{
  ...
  entensions: {
    "revocation_endpoint": "https://server.example.com/connect/revoke"
  }
}
```
?

ericchiang:
To the best of my knowledge there's no easy way. Will look into it.

bobbyrullo:
@ericchiang @yutongp - can one of you explain here what makes the marshalling so tricky here?

yutongp:
As I explained, ideally `revocation_endpoint` should not be a field in `extensions`, but in root level with `token_endpoint`, `authorization_endpoint`.

bobbyrullo:
Ah I see. Given that these *are* extensions I'm actually ok with it being in "extensions" regardless what others have done; if it ever makes it into the standard we can migrate it into the root level...

ericchiang:
@yutongp for your use case are you acting as the client or the provider? e.g. Are you attempting to get the revocation_endpoint or server it?

ericchiang:
the returned provider now has a method for accessing any non-standard fields https://godoc.org/github.com/coreos/go-oidc#Provider.Claims

PR #56, opened 2016-02-25
Do not convert Unix timestamps to float64

### LINKED ISSUE #55: oidc: floating point values in claims don't interop well with certain OIDC implementations
According to the [OpenID Connect specification section 2. ID Token](http://openid.net/specs/openid-connect-core-1_0.html#IDToken) the ``exp`` and ``iat`` claims are REQUIRED and specified as JSON number representing the number of seconds from 1970-01-01T0:0:0Z.

The current implementation converts the timestamps from golang Time, to Unix int64 and then finally converts that to a float64. When doing this the numbers come across as ``1.456478332e+09`` instead of ``1456478332``.

Other implementations, like mod_auth_openidc for Apache used by among others OpenStack, does not understand Unix timestamps represented as 64-bit floating point numbers.

I would argue that it breaks the spec by using float64, even though it does not explicitly say "integer" rathern than just JSON numbers.

Is there any reason for why the 64-bit integers are converted to 64-bit floating point numbers?

fnordahl:
Fixes #55

### DIFF
--- a/jose/claims.go
+++ b/jose/claims.go
@@ -3,6 +3,7 @@ package jose
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


ericchiang:
@bobbyrullo @sym3tri if you have comments on this please let us know. This is back from the initial go-oidc commit so there's no context for the existing code.

fnordahl:
I will look over the failing tests and suggest appropriate fixes this afternoon.

ericchiang:
Pretty small issue. The test case just uses float64 encoding

```
--- FAIL: TestNewClaims (0.00s)
	util_test.go:93: want=jose.Claims{"iss":"https://example.com", "sub":"user-123", "aud":"client-abc", "iat":-6.21040608e+10, "exp":-6.21040572e+10} got=jose.Claims{"sub":"user-123", "aud":"client-abc", "iat":-62104060800, "exp":-62104057200, "iss":"https://example.com"}
```

bobbyrullo:
LGTM

Issue #54, opened 2016-02-25
basic auth uses wrong encoding scheme


ericchiang:
Brought up in coreos/dex#336. The OAuth2 spec requires the client ID and secret to be URL encoded before being used in base auth.

> Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server.  __The client identifier is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm__ per
   Appendix B, __and the encoded value is used as the username; the client
   password is encoded using the same algorithm and used as the
   password.__

PR #58, opened 2016-02-29
basic auth with URL encode

gerson24:
The OAuth2 spec requires the client ID and secret to be URL encoded before being used in base auth.

Fixes: #54


ericchiang:
Can you write a test? :)

gerson24:
do you like it this way?

PR #59, opened 2016-02-29
http: remove HandlerClient and RequestRecorder

### LINKED ISSUE #57: go-oidc/http package pulls in test-only net/http/httptest
The go `net/http/httptest` package should only be used in test builds. In particular, that package adds the `httptest.serve` flag which leaks into binaries depending on go-oidc (e.g. https://github.com/kubernetes/kubernetes/issues/21114)

From what I can tell, the [`HandlerClient`](https://github.com/coreos/go-oidc/blob/46fc3c20dcad27d27596c8832b354621bdd8b88f/http/client.go#L13) that uses `httptest` is only used in testing, so I suggest moving it to a test-only package (such as `go-oidc/http/testing` or `go-oidc/http/httptest`)

ericchiang:
Users should just use net/http/httptest instead.

Closes #57

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -1,51 +1,7 @@
 package http
 
-import (
-	"io/ioutil"
-	"net/http"
-	"net/http/httptest"
-)
+import "net/http"
 
 type Client interface {
 	Do(*http.Request) (*http.Response, error)
-}
-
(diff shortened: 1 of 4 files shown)


bobbyrullo:
LGTM

Issue #50, opened 2016-02-22
Replace jose package with go-jose


bsr203:
Hi,

This is not an issue, but just to inform a library I found, which may be interest to you

https://github.com/square/go-jose

Looks well maintained and comprehensive.

Cheers,
bsr

gtank:
It's more feature-complete and might have had more eyes on it. Worth investigating, since it would let us move away from PKCS1v15 RSA.

ericchiang:
Spoke with @gtank about this offline.

Our current jose implementation has a lot of missing features (support for ecdsa Keys), as well as documentation issues (for instance we've had internal issues with developers not correctly verifying JWTs).

While migrating isn't a top priority, in the long term I think it's best for us to deprecate the jose package and move to square/go-jose.

ericchiang:
done https://github.com/coreos/go-oidc/pull/110

PR #61, opened 2016-03-03
Add simpler OpenID Connect client 

ericchiang:
This PR adds a OpenID Connect client under the path "github.com/coreos/go-oidc"

Large highlights are:

* Works with existing golang.org/x/oauth2 package.
* Moves to github.com/square/go-jose for JWT validation.

This is not indented to replace any of the existing packages for server side code, but presents a much simpler interface for clients. It tries to mimic the idioms of golang.org/x/oauth2 package as much as possible and should be much easier for users who only want to worry about client side code.

EDIT: Example program [here](https://github.com/ericchiang/go-oidc/blob/new_client/example/app.go)


bobbyrullo:
This is nice work, but I think it's confusing to have two OIDC clients in the same package. Maybe this should be it's own thing?

ericchiang:
I'd be happy to move this to another repo :)

Would it be okay to deprecate the current client? Overall it seems like we're going to need to address this for other packages in this repo:

* http: not a package we should export
* jose: users should use square/go-jose (#50)
* oauth2: users should use golang.org/x/oauth2 (#64)
* key: for JWKs users should use square/go-jose, other logic (e.g. PrivateKeyRotator) is primarily geared towards dex.
* oidc: this is my attempt to clean up the client logic, the server side is missing the oauth2 server parts (which are in dex)

At some point I think it's more reasonable for us to move the dex parts into dex, and present a cleaner interface for package users.

bobbyrullo:
> I'd be happy to move this to another repo :)

Nah, let's keep it here.

> Would it be okay to deprecate the current client? Overall it seems like we're going to need to address this for other packages in this repo:

This is such radical, I'd rather just call this a 2.0.

> * http: not a package we should export

Yeah, agreed - I don't see much value here except for testing.

> * jose: users should use square/go-jose (#50)

+1

> * oauth2: users should use golang.org/x/oauth2 (#64)

+1

> * key: for JWKs users should use square/go-jose, other logic (e.g. PrivateKeyRotator) is primarily geared towards dex.

Yeah, the privatekeyrotator shouldn't be in here.

> * oidc: this is my attempt to clean up the client logic, the server side is missing the oauth2 server parts (which are in dex)

> At some point I think it's more reasonable for us to move the dex parts into dex, and present a cleaner interface for package users.

If there are generally useful parts for implementing OIDC servers, we can keep them here.

Maybe the package structure should be "client" and "server" instead of oidc, key, etc.

ericchiang:
> This is such radical, I'd rather just call this a 2.0.

There are already a few things that depend on go-oidc. I've never had fun when trying to resolve package name changes with dependency evaluation tools like godep. It's one thing to change a few signatures, it's another thing to delete, move, or remove a package.

We might consider using http://labix.org/gopkg.in style vanity path and just work off a branch. Though I don't like the idea of yet another site being between a `go get` and our code.

> If there are generally useful parts for implementing OIDC servers, we can keep them here.

> Maybe the package structure should be "client" and "server" instead of oidc, key, etc.

Any parts you want to keep for the server side? Since adding a new package "server" would require users to migrate anyway, I think it may be easier to leave the existing code for now and add it later if there's an interest. (Looks like polvi is the only one using go-oidc publicly https://godoc.org/github.com/coreos/go-oidc/oidc?importers)

Additional notes from IRL conversations:

* Need to add claims assertions for JWT validators. e.g. "issuer must be this" which is common for ID Tokens. Should we export this functionality or reduce API scope and do it internally?

NOTE: Will add some use cases for this. Have been using the code from the [remote key set](https://github.com/ericchiang/go-oidc/blob/new_client/remote_jwks.go) for a lot of things lately.

ericchiang:
cc @dghubble feel free to comment if you have any feedback. I know we've discussed this for gologin

client.go
```diff
@@ -0,0 +1,178 @@
...
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	var userinfo struct {
```
dghubble: why not re-use the UserInfo you defined above? `userinfo := new(UserInfo)`

client.go
```diff
@@ -0,0 +1,178 @@
...
+	if u.rawClaims != nil {
+		return u.rawClaims[key]
+	}
+	return nil
+}
+
+// UserInfo uses the token source to query the provider's userinfo endpoint.
+func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource) (*UserInfo, error) {
+	if p.UserInfoURL == "" {
+		return nil, errors.New("oidc: provider doid not provide a userinfo endpoint")
```
dghubble: doid?

client.go
```diff
@@ -0,0 +1,178 @@
...
+	return &p, nil
+}
+
+// Endpoint returns the OAuth2 auth and token endpoints for the given provider.
+func (p *Provider) Endpoint() oauth2.Endpoint {
+	return oauth2.Endpoint{AuthURL: p.AuthURL, TokenURL: p.TokenURL}
+}
+
+// UserInfo represents the OpenID Connect userinfo claims.
+type UserInfo struct {
```
dghubble: nit: can this be moved up or down so all the `(p *Provider)` methods are together?

doc.go
```diff
@@ -0,0 +1,63 @@
+/*
+Package oidc implements OpenID Connect client logic for the golang.org/x/oauth2 package.
+
+	provider, err := oidc.NewProvider(ctx, "https://accounts.exmaple.com")
+	if err != nil {
+		return err
+	}
+
+	// Configure an OpenID Connect aware OAuth2 client.
```
dghubble: All that is needed to make an OAuth2 client OpenID Connect aware is providing the right endpoint and scopes right? Just checkin

ericchiang: Yep!

test
```diff
@@ -14,9 +14,7 @@ COVER=${COVER:-"-cover"}
 
 RACE=${RACE:-"-race"}
 
-source ./build
-
-TESTABLE="http jose key oauth2 oidc"
+TESTABLE="http jose key oauth2 oidc ."
```
dghubble: I recently switched to `PKGS=$(go list ./... | grep -v /vendor)` instead of listing packages

doc.go
```diff
@@ -0,0 +1,63 @@
...
+
+		// ...
+	})
+
+This package uses contexts to derive HTTP clients in the same way as the oauth2 package. To configure
+a custom client, use the oauth2 packages HTTPClient context key when constructing the context.
+
+	myClient := &http.Client{}
+
+	myCtx := context.WithValue(parentCtx, oauth2.HTTPClient, myClient)
```
dghubble: Super handy for writing tests too! Allows higher level libs to create test clients, its a big win. https://github.com/dghubble/gologin/blob/master/github/login_test.go#L25

dghubble:
Yes! Yes! Yes!

bobbyrullo:
Heres the plan going forward:

- [ ] Version the current repo
- [ ] Create a new branch for OIDC v2
- [ ] Incremental Changes replacing OAuth2, Jose
- [ ] Set up vanity URLs for v1

ericchiang:
Note that `go get gopkg.in/coreos/go-oidc.v2` will pull from the `v2` branch. We can start working off that everything will just work.

ericchiang:
@bobbyrullo @sym3tri there's another practical issue with this plan. dex uses this package a lot. We aren't going to be able to rework this package without a large overhaul of dex.

```
$ grep -r -E -o -h '(oidc|jose|oauth2)\.[[:alnum:]]+' $(find . -name *.go | grep -v 'vendor') | sort | uniq -c | sort -nr
     71 oidc.ClientIdentity
     71 oidc.ClientCredentials
     69 oauth2.NewError
     45 oidc.Identity
     41 oidc.ClientMetadata
     28 oauth2.ErrorServerError
     27 jose.Claims
     24 oauth2.ErrorInvalidRequest
     15 oauth2.ErrorInvalidClient
     15 jose.NewSignedJWT
     14 oidc.LoginFunc
     14 jose.Signer
     11 oidc.ProviderConfig
     10 oidc.NewClaims
      9 oidc.Client
      9 jose.JWK
      8 oidc.ClientRegistrationResponse
      8 oauth2.ErrorUnsupportedResponseType
      7 oidc.NewClient
      7 oidc.ClientConfig
      7 oauth2.GrantTypeAuthCode
      7 jose.JWT
      6 oauth2.Client
      6 jose.ParseJWT
      5 oidc.FetchProviderConfig
      5 oauth2.ErrorAccessDenied
      4 oidc.NewJWTVerifier
      4 oidc.GenClientID
      4 oauth2.ErrorInvalidGrant
      4 oauth2.Error
      3 oidc.JWTVerifier
      3 oidc.ExtractBearerToken
      3 oauth2.TokenResponse
      3 oauth2.GrantTypeClientCreds
      2 oidc.ClientCredsTokenRefresher
      2 oidc.AuthenticatedTransport
      2 oauth2.NewClient
      2 oauth2.ErrorUnsupportedGrantType
      2 oauth2.Config
      2 oauth2.ClientCredentials
      2 oauth2.AuthMethodClientSecretPost
      1 oidc.VerifySignature
      1 oidc.VerifyClientClaims
      1 oidc.IdentityFromClaims
      1 oauth2.ResponseTypeCode
      1 oauth2.GrantTypeRefreshToken
```

PR #65, opened 2016-03-12
jose/oidc fixes for non-comformity

saracen:
**JOSE**
JWK's "kty" parameter is case-sensitive:
https://tools.ietf.org/html/rfc7517#section-4.1

The "alg" parameter is case-sensitive and optional:
https://tools.ietf.org/html/rfc7517#section-4.4

Given that it is optional, I think it's safe to assume that there should be a default (the specification has a "recommended" algorithm for each "kty"). This appears to be the behaviour with other clients and servers.

**OIDC**
The specification only states that jwks_uri SHOULD have a Cache-Control header:
http://openid.net/specs/openid-connect-core-1_0.html#rfc.section.10.2.1

### DIFF
--- a/jose/sig.go
+++ b/jose/sig.go
@@ -2,7 +2,6 @@ package jose
 
 import (
 	"fmt"
-	"strings"
 )
 
 type Verifier interface {
@@ -17,7 +16,7 @@ type Signer interface {
 }
 
 func NewVerifier(jwk JWK) (Verifier, error) {
-	if strings.ToUpper(jwk.Type) != "RSA" {
+	if jwk.Type != "RSA" {
 		return nil, fmt.Errorf("unsupported key type %q", jwk.Type)
 	}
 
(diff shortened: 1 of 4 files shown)


saracen:
I probably should have added: I made these changes because they're all required to get Azure AD working with this client.

bobbyrullo:
saracen - thanks for the PR! Just a minor doc request, and then LGTM.

saracen:
Updated documentation :)

It would be nice to add this soon, but, some other thoughts that could be implemented later:

- If no TTL is provided for the keys we could refresh at the same time as when we re-sync the OP config?
- Is there any value in NewRemotePublicKeyRepo being public?

bobbyrullo:
LGTM

bobbyrullo:
@saracen - the answers to those are "probably" and "probably not". They both seem fairly uncontroversial, so feel free to open a couple of PRs (i.e. no proposal, issue needed) but if you're not going to do the work, just open up issues and we'll get to it at some point.

PR #67, opened 2016-03-22
Fix #66 by trimming trailing slashes from discovery, issuer url.

### LINKED ISSUE #66: Handle trailing slash in Auth0 openid-configuration issuer
A bit of a snafu with FetchProviderConfig...

https://fullung.auth0.com/.well-known/openid-configuration has issuer:https://fullung.auth0.com/ but https://fullung.auth0.com//.well-known/openid-configuration (double slash intentional) is a 404.

This is a problem because provider.go simply does:

req, err := http.NewRequest("GET", r.issuerURL+discoveryConfigPath, nil)

and later:

if !urlEqual(cfg.Issuer.String(), r.issuerURL)

Probably want to change provider.go to do:

if !urlEqual(cfg.Issuer.String(), r.issuerURL) && !urlEqual(cfg.Issuer.String(), r.issuerURL+"/") {

Not sure that getting Auth0 to fix their endpoint is going to happen and there might be other OpenID providers out there that made the same mistake.

Thoughts appreciated.

alberts:


oidc/provider.go
```diff
@@ -644,7 +650,7 @@ func (r *httpProviderConfigGetter) Get() (cfg ProviderConfig, err error) {
 
 	// The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve the configuration information.
 	// http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation
-	if !urlEqual(cfg.Issuer.String(), r.issuerURL) {
+	if !urlEqual(strings.TrimSuffix(cfg.Issuer.String(), "/"), r.issuerURL) {
```
ericchiang: Why trim the suffix here? Shouldn't this be exactly equal? E.g. if r.issuerURL is "https://foo/auth/" then the returned issuer should be exactly "https://foo/auth/" as well.

alberts:
ptal

ericchiang:
LGTM

PR #68, opened 2016-03-30
Fixed PublicKey JSON Marshaling

Quentin-M:
Because the `MarshalJSON()` method implemented by `jose/JWK` has a pointer receiver, we must pass a pointer to it. Otherwise, the method is never called when a `key/PublicKey` instance is marshaled, the Go default MarshalJSON function is used on `jose/JWK`. Consequently, the JSON data for a PublicKey is invalid (i.e. modulus and exponent are not properly encoded) and it is impossible to unmarshal it back to a valid PublicKey.

### DIFF
--- a/key/key.go
+++ b/key/key.go
@@ -20,7 +20,7 @@ type PublicKey struct {
 }
 
 func (k *PublicKey) MarshalJSON() ([]byte, error) {
-	return json.Marshal(k.jwk)
+	return json.Marshal(&k.jwk)
 }
 
 func (k *PublicKey) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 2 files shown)


ericchiang:
Actually maybe we just want to define MarshalJSON on the concrete type of `jose.JWK` and `key.PublicKey` rather than the pointer.

```go
func (j JWK) MarshalJSON() ([]byte, error) {
	t := jwkJSON{...}
	return json.Marshal(&t)
}
```

Instead of

```go
func (j *JWK) MarshalJSON() ([]byte, error) {
	t := jwkJSON{...}
	return json.Marshal(&t)
}
```

This would make both mashaling a pointer to JWK and a concrete value work.

```go
jwk := jose.JWK{}
json.Marshal(jwk)
json.Marshal(&jwk) // works even if MarshalJSON is defined on the concrete type.
```

See example here: https://play.golang.org/p/fv5MFSk15X

ericchiang:
Anyway, LTGM. I've opened a different issue for the MarshalJSON stuff.

Quentin-M:
Thanks!

Issue #64, opened 2016-03-11
Proposal: replace oauth2 with golang.org/x/oauth2


ericchiang:
In the same vein as #50. Go's oauth2 package is better maintained and has been tested against a wide array of oauth2 providers. We already have to ensure that dex works with this client, is there a reason against using it?

ericchiang:
Done https://github.com/coreos/go-oidc/pull/110

PR #70, opened 2016-04-08
oauth2: return embedded http client

srenatus:
This is a prerequisite to a fix for https://github.com/coreos/dex/issues/395.

Happy to add tests if you agree this is potentially mergeable.

(Things will look different with #64, but I'd be happy to help transitioning, too.)

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -140,6 +140,11 @@ func NewClient(hc phttp.Client, cfg Config) (c *Client, err error) {
 	return
 }
 
+// Return the embedded HTTP client
+func (c *Client) HttpClient() phttp.Client {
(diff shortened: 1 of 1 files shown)


srenatus:
Could this be some rate-limit issue? `go get golang.org/x/tools/cmd/vet` works fine from here...

ericchiang:
You can already set the http client through the [client config](https://godoc.org/github.com/coreos/go-oidc/oidc#ClientConfig). Plus this doesn't set the http client, it only returns it.

Is this trying to accomplish something else?

bobbyrullo:
he's trying to *get* the client config not set it. see #395

bobbyrullo:
LGTM

PR #71, opened 2016-04-08
update go in travis

srenatus:
Not sure what happened in #70's travis run, but `1.6` is missing there. Not sure about also dropping `1.4`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ language: go
 
 go:
   - 1.4.3
-  - 1.5.2
+  - 1.5.3
+  - 1.6
 
 install:
  - go get -v -t ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ go:
 install:
  - go get -v -t ./...
  - go get golang.org/x/tools/cmd/cover
- - go get golang.org/x/tools/cmd/vet
 
 script:
  - ./test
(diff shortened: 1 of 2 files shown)


ericchiang:
We need to support `1.4` since this is used by projects do support it.

There are two things we need to do for travis.
* Remove the `go get golang.org/x/tools/cmd/vet` command in the same vein as coreos/dex#397.
* If testing with `1.4`, don't execute [this block](https://github.com/coreos/go-oidc/blob/f427f54ef96beaa1db5fabbe4fff00de535d5494/test#L51-L57) in the test script.

Would you be okay updating this PR with those changes?

srenatus:
@ericchiang sure. Updated. Let's see what travis has to say about it :grinning:

srenatus:
So Travis is happy. There will probably be a reason to touch those files again [the day after tomorrow (Go 1.5.4, 1.6.1)](https://groups.google.com/forum/#!topic/golang-nuts/MmSbFHLPo8g), though. Not sure how fast Travis will pick those up...

ericchiang:
> Not sure how fast Travis will pick those up...

Travis just uses [gvm](https://github.com/moovweb/gvm) if I recall correctly which just downloads them from https://golang.org/dl/. So it should catch it as soon as it's out.

ericchiang:
lgtm, thanks!

srenatus:
@ericchiang: thanks for the hint re: travis and go, I wasn't aware of this.

PR #76, opened 2016-04-20
*: decode iat and exp as int64

ericchiang:
Since #56 this package encodes the 'iat' and 'exp' claims as int64
instead of float64. However it still decodes them as a float64.
This causes comparisions to fail in tests that use some version of
reflect.DeepEqual. See coreos/dex#418.

To avoid this, attempt to decode the 'iat' and 'exp' explicitly as
int64 if possible.

cc @sym3tri @bobbyrullo

### DIFF
--- a/jose/claims.go
+++ b/jose/claims.go
@@ -105,6 +105,25 @@ func decodeClaims(payload []byte) (Claims, error) {
 	if err := json.Unmarshal(payload, &c); err != nil {
 		return nil, fmt.Errorf("malformed JWT claims, unable to decode: %v", err)
 	}
+
+	// Since we encode these claims as int64, it's advantageous for comparision to
+	// decode them as the same type.
+	//
+	// See https://github.com/coreos/dex/issues/418
+	var stdClaims struct {
+		IssuedAt int64 `json:"iat"`
+		Expires  int64 `json:"exp"`
+	}
+
+	// No error case. If decoding fails continue.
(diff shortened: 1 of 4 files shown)


bobbyrullo:
This goes against the spec IMO (see my comment here: https://github.com/coreos/go-oidc/issues/55#issuecomment-189404209)

It's ok for us to encode an int, since ints are also floats in JSON. But people using this library may be parsing claims generated elsewhere, which may be floats.

ericchiang:
> This goes against the spec IMO (see my comment here: #55 (comment))

Yes I'm aware that JSON numeric types are considered float64 by default. But practically encoding/json can tell the difference and `reflect.DeepEqual` will complain. The spec doesn't say what kind of types the Go structs must hold when decoding JSON values.

I don't like this solution, but there doesn't seem like a better one so long as we represent claims as an arbitrary map of `string`s to `interface{}`s.

> It's ok for us to encode an int, since ints are also floats in JSON. But people using this library may be parsing claims generated elsewhere, which may be floats.

As the decoding logic currently stands with this change:

1. First, decode all claims
2. If 'iat' and 'exp' can be decoded as ints, use those values instead of floats
3. If they can't be, use whatever values were parsed in step 1.

This works perfectly fine with implementations that use float64 values.

ericchiang:
closing in favor of solutions in https://github.com/coreos/dex/issues/418#issue-149532536

PR #78, opened 2016-04-20
adding user_credentials grant type 

gambol99:
- adding the user_credentials grant type

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -56,6 +56,7 @@ const (
 const (
 	GrantTypeAuthCode     = "authorization_code"
 	GrantTypeClientCreds  = "client_credentials"
+	GrantTypeUserCreds    = "password"
 	GrantTypeImplicit     = "implicit"
 	GrantTypeRefreshToken = "refresh_token"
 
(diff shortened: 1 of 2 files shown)


ericchiang:
As a note, grant_type=password is an OAuth2 feature not used by OpenID Connect. Are you using the oauth2 package on it's own, or are you attempting to interact with an OpenID Connect provider?

gambol99:
Apologizes, I wasn't aware of that. Were using library to interact with [Keycloak](http://keycloak.jboss.org/) service.

oauth2/oauth2.go
```diff
@@ -225,6 +226,30 @@ func (c *Client) ClientCredsToken(scope []string) (result TokenResponse, err err
 	return parseTokenResponse(resp)
 }
 
+// UserCredsToken posts the username, password, client_id and optional client_secret depending if client is public or not.
+// May not be supported by all OAuth2 servers.
+func (c *Client) UserCredsToken(username, password string, scope []string) (result TokenResponse, err error) {
```
ericchiang: Scopes are already on the Client struct, please use those.

gambol99:
@ericchiang ... removed the param and using the c.scope

ericchiang:
lgtm

PR #81, opened 2016-05-03
oidc: Export JWT in AuthenticatedTransport

bobbyrullo:
This support the case in which you have a Token already that may or may
not be invalidated.

### DIFF
--- a/oidc/transport.go
+++ b/oidc/transport.go
@@ -67,6 +67,15 @@ func (t *AuthenticatedTransport) verifiedJWT() (jose.JWT, error) {
 	return t.jwt, nil
 }
 
+// SetJWT sets the JWT held by the Transport.
+// This is useful for cases in which you want to set an initial JWT.
+func (t *AuthenticatedTransport) SetJWT(jwt jose.JWT) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.jwt = jwt
+}
+
 func (t *AuthenticatedTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	jwt, err := t.verifiedJWT()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


oidc/transport.go
```diff
@@ -47,24 +47,24 @@ type AuthenticatedTransport struct {
 	http.RoundTripper
 
 	mu  sync.Mutex
```
ericchiang: Don't we also have to export the mu as well then?

Maybe:

```
func (r *AuthenticatedTransport) JWT() jose.JWT
```

bobbyrullo: :( good point. But that doesn't help me - I need to set an initial JWT. Could add a getter and setter I guess, which seems the safest.

bobbyrullo: would that work for you eric?

ericchiang: Comment from slack:

TokenRefresher is an interface. why not this?

```
type tokenRefreshWithInitialJWT struct {
    jwt       jose.JWT
    once      sync.Once
    refresher oidc.TokenRefresher
}

func (t *tokenRefreshWithInitialJWT) Verify(jwt jose.JWT) error {
    return t.refresher.Verify(jwt)
}

func (t *tokenRefreshWithInitialJWT) Refresh() (jwt jose.JWT, err error) {
    // Only set the initial token once.
    t.once.Do(func() { jwt = t.jwt })
    if jwt != nil {
        return jwt, nil
    }
    return t.refresher.Refresh()
}
```

getter/setter works fine too if you feel it's a common enough use case

or maybe a constructor that sets the private fields

bobbyrullo: Note that this is for AuthenticatedTransport not TokenRefresher.

ericchiang: Right. The AuthenticatedTransport takes a TokenRefresher which you can
customize to fit your needs.

bobbyrullo: Ah. I think your solution is more complicated. It's also less generally useful - my way would work for an AuthenticatedTransport with any TokenRefresher

ericchiang: Sure. Yeah, getter and setter works for me.

bobbyrullo: PTAL

ericchiang:
Lgtm

bobbyrullo:
Thanks!

Issue #14, opened 2015-08-17
Add constants and methods for accessing common claims


bobbyrullo:
eg. 

```
ClaimsSub = "sub"
...
func (c *Claims) GetSub() string
```

PR #84, opened 2016-05-11
oidc: Add an attribute to the Identity Entity for managing additional…

### LINKED ISSUE #83: oidc.Identity is too limited; does not allow for setting additional attributes related to Identity.
oidc.Identity is too limited in functionality.  Without replacing it as a type, users of go-oidc not able to add additional attributes to an Identity.

I need to modify dex to add Additional Claims to the JWT.  To accomplish this, I replaced the oidc.Identity type in dex with a new dex specific type.  The result turned out to be oidc.Identity with one additional field.  

I don't see any advantages to doing the mechanical work of replacing the type within dex.  I also assume any other users of the go-oidc package will eventually run into a similar issue.  So rather than continue down the path of looking at replacing oidc.Identity in dex, I would like to make the existing type less limited.

clc-github:
… claims.

This change allows the Identity Entity to have Additional Claims
attrbutes set as allowed in section 5.1.2 of OpenID Connect Core
1.0.

This change will enable downstream users, such as dex, to use
additional claims without resorting to creating a replacement or
wrapping oidc.Identity.

Fixes #83

### DIFF
--- a/oidc/identity.go
+++ b/oidc/identity.go
@@ -41,17 +41,26 @@ func IdentityFromClaims(claims jose.Claims) (*Identity, error) {
 		ident.ExpiresAt = exp
 	}
 
+	additonalClaims := jose.Claims{}
+	for k, v := range claims {
+		if jose.IsAdditonalClaim(k) {
+			additonalClaims.Add(k, v)
+		}
+	}
+
+	// Don't retain a reference if no Additional Claims are present.
+	if len(additonalClaims) > 0 {
+		ident.AdditonalClaims = additonalClaims
+	}
+
 	return &ident, nil
 }
 
 // CopyAdditonalClaims copies all additional claims to claims
 func (ident *Identity) CopyAdditonalClaims(claims jose.Claims) {
-	if ident.AdditonalClaims == nil {
-		return
-	}
-
 	for k, v := range ident.AdditonalClaims {
-		// TODO: ignore standard claims; related to issue #14
-		claims.Add(k, v)
+		if jose.IsAdditonalClaim(k) {
+			claims.Add(k, v)
+		}
 	}
 }
--- a/oidc/identity.go
+++ b/oidc/identity.go
@@ -8,10 +8,11 @@ import (
 )
 
 type Identity struct {
-	ID        string
-	Name      string
-	Email     string
-	ExpiresAt time.Time
+	ID              string
+	Name            string
+	Email           string
+	ExpiresAt       time.Time
+	AdditonalClaims jose.Claims
 }
 
 func IdentityFromClaims(claims jose.Claims) (*Identity, error) {
@@ -41,4 +42,16 @@ func IdentityFromClaims(claims jose.Claims) (*Identity, error) {
 	}
 
 	return &ident, nil
+}
+
+// CopyAdditonalClaims copies all additional claims to claims
+func (ident *Identity) CopyAdditonalClaims(claims jose.Claims) {
+	if ident.AdditonalClaims == nil {
+		return
+	}
+
+	for k, v := range ident.AdditonalClaims {
+		// TODO: ignore standard claims; related to issue #14
+		claims.Add(k, v)
+	}
 }
--- a/oidc/identity.go
+++ b/oidc/identity.go
@@ -8,11 +8,11 @@ import (
 )
 
 type Identity struct {
-	ID              string
-	Name            string
-	Email           string
-	ExpiresAt       time.Time
-	AdditonalClaims jose.Claims
+	ID          string
+	Name        string
+	Email       string
+	ExpiresAt   time.Time
+	ExtraClaims jose.Claims
 }
 
 func IdentityFromClaims(claims jose.Claims) (*Identity, error) {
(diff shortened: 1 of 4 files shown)


oidc/identity.go
```diff
@@ -42,3 +43,15 @@ func IdentityFromClaims(claims jose.Claims) (*Identity, error) {
 
 	return &ident, nil
 }
+
+// CopyAdditonalClaims copies all additional claims to claims
+func (ident *Identity) CopyAdditonalClaims(claims jose.Claims) {
```
ericchiang: Can you elaborate more on why you want this to be a method we expose as part of the package API? It seems unnecessary.

oidc/identity.go
```diff
@@ -8,10 +8,11 @@ import (
...
-	Email     string
-	ExpiresAt time.Time
+	ID              string
+	Name            string
+	Email           string
+	ExpiresAt       time.Time
+	AdditonalClaims jose.Claims
 }
 
 func IdentityFromClaims(claims jose.Claims) (*Identity, error) {
```
ericchiang: I'd assume that `IdentityFromClaims` would also fill the `AdditionalClaims` field

oidc/identity.go
clc-github: 1. So additional sanity checks can be added, such as preventing Standard Claims from being copied unintentionally.  This assumes issue #14 is partially addressed so the full list of Standard Claims is known.
1. To prevent downstream systems, such as dex, from each having to implement something similar.

My goal and the origin of this functionality is to be able to copy Additional Claims in a dex connector in 'server/server.go' right before the `user.AddToClaims(claims)` calls.

oidc/identity.go
clc-github: I intend to add that functionality.  It depends on some of the functionality needed for issue #14.  I'm preparing another PR that would address this.

oidc/identity.go
ericchiang: Right, but even the nil check is unnecessary. You can range over a nil map.

```go
package main

import (
	"fmt"
)

func main() {
	m1 := map[string]string{}
	var m2 map[string]string // m2 is nil

	for key, val := range m2 {
		m1[key] = val
	}
	fmt.Println(m1)
}
```

oidc/identity.go
ericchiang: Please add it to this PR.

oidc/identity.go
clc-github: Yes this check is not needed in this version, I will remove it.

oidc/identity.go
clc-github: ok

oidc/identity.go
ericchiang: My point is that it's not something we need to expand the package API for.

clc-github: If you have the point of view that `AdditionalClaims` can or should contain all Claims not in the struct, then I agree.  I was attempting to have it contain only Claims that dex (and others) were unlikely to use in an attempt to avoid any accidental interference.

 https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
 https://tools.ietf.org/html/rfc7519#section-4.1
 https://openid.net/specs/openid-connect-core-1_0.html#IDToken

I will update the PR such that AdditionalClaims will contain all claims not otherwise included in the struct.

ericchiang:
@clc-github I've got some further inline comments on one of the intermediate commits regarding adding `IsStandardClaim` and `IsAdditonalClaim` [here](https://github.com/clc-github/go-oidc/commit/c7a630dfca2aace09c4c186d5ce330f5ee744114#comments).

My overall position is, despite the spec's claim a majority of these fields aren't particularly standard for actual OpenID Connect implementations, and including them seems like more like a direct translation of the spec rather than providing useful utilities around OIDC.

Having an enumeration of claims defined by OpenID Connect is fine. And adding AdditionalClaims (or just calling it "RawClaims" or "ExtraClaims") to the Identity struct is fine as well. But I don't think we need the additional functions.

Does this sound reasonable?

clc-github:
@ericchiang, I have a different position, but my pov is one of an integrator that needs a sanity check to make sure no unintentional claims are introduced from non-dex code that might overwrite a Claim that dex (or go-oidc) is responsible for managing.

I will remove the additional functions and constants, and rename AdditionalClaims to ExtraClaims.

ericchiang:
Cool. I will note that there's no guarantee that dex only deals with standard claims ATM. If you are concerned with conflicts for, say, LDAP attributes, I'd recommend "namespacing" the attribute names with something like "https://coreos.com/ldap/" for example https://github.com/coreos/dex/blob/b29deb736b160304d072dd0a54e8a7c504bad3c4/user/user.go#L26

ericchiang:
LGTM, but can @fnordahl and/or @bobbyrullo also sign off on this?

fnordahl:
Nice. This would make it possible to do further processing of claims in Dex. 👍 

We should probably also add some code to verify/marshall/unmarshall the standard claims in Dex then. But let's get to that as soon as this one is merged.

I guess this also will put pressure on making Dex's tests play nicely with my float64 -> int64 changes to go-oidc. I will try harder to make that one work and fix it ASAP.

I still think that the standard claims really should be handled by the OIDC library, but i'll let that remain my opinion and stand behind your decision to make it up to Dex to handle them.

ericchiang:
Sure. And I'm happy adding standard claims to the identity object. My main objection for this PR was exposing an `IsStandardClaim` method which I don't think is as useful.

ericchiang:
The new client only parses required claims and exposes all other claims to the user as they wish. It's up to the user to determine what's standard.

Please take a look at that https://godoc.org/github.com/coreos/go-oidc#Provider.Claims

PR #85, opened 2016-05-12
removing the debug log

gambol99:
- changing from info to debug level on the provider config as it more debug and required information. On similar note, it would be nice to allow the consumer to set the io.Writer on the logs?


bobbyrullo:
LGTM - as for you other request...we're in the middle of re-evaluating what logging lib we're using here at coreos generally, so....stay tuned :)

PR #86, opened 2016-05-15
oidc: Add Standard Claims and Additional Claims

fnordahl:
Extend data structures to hold all the Standard Claims per
http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

Add data structures to handle Additional Claims

Implement supporting functions to marshall and unmarshall the data

### DIFF
--- a/oidc/identity.go
+++ b/oidc/identity.go
@@ -7,11 +7,36 @@ import (
 	"github.com/coreos/go-oidc/jose"
 )
 
-type Identity struct {
-	ID        string
-	Name      string
-	Email     string
-	ExpiresAt time.Time
+type Address struct { // 5.1.1 Address Claim
+	Formatted     string // formatted
+	StreetAddress string // street_address
+	Locality      string // locality
+	Region        string // region
+	PostalCode    string // postal_code
+	Country       string // country
+}
+
+type Identity struct { // 5.1. Standard Claims
+	ID                  string // sub
+	Name                string // name
+	GivenName           string // given_name
+	FamilyName          string // family_name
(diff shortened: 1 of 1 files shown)


fnordahl:
This PR is an Request For Comments at this stage, please add the "do not merge" label for now.

During my on-going work to add support for groups to the LDAP Connector (see coreos/dex#432) I have come across a need to extend the data structures of the go-oidc library.

Given that there has been some discussion back and forth wether this should be a Dex specific type or if it belongs to the go-oidc library on the mailing lists; I ask before I continue work on this approach. My opinion is that it definetively should be in the go-oidc library as the claims are a part of the OpenID Connect specification.

I would love to have some feedback and discussion on this matter! @bobbyrullo @ericchiang

ericchiang:
See conversation here https://github.com/coreos/go-oidc/issues/83#issuecomment-218933353 and here https://github.com/coreos/go-oidc/pull/84

Bottom line, dex should move away from using structs in go-oidc for storing data. That's a dex implementation detail that shouldn't keep bleeding back into go-oidc. We're already having trouble updating go-oidc in dex https://github.com/coreos/dex/issues/418, and __if we ever want a claim that's not standard OpenID Connect we'll have to create a separate struct in dex anyway__.

So yeah, I'm for the dex specific type.

As for chances to go-oidc regarding standard and extra claims, please see https://github.com/coreos/go-oidc/pull/84

fnordahl:
Thank you for the swift response and references to the other discussions. I guess I should do better research before firing off PR's the next time. Closing this one for now.

PR #87, opened 2016-06-10
oauth2: add support for Azure AD

tksm:
This change adds support for Azure AD token response on Code Flow.

Though RFC6749 section 5.1 states Numerical values are included as
JSON numbers, Azure AD returns "expires_in" value as string.

see Azure AD JSON example
https://msdn.microsoft.com/en-us//library/azure/dn645542.aspx

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -332,16 +332,16 @@ func parseTokenResponse(resp *http.Response) (result TokenResponse, err error) {
 		result.Scope = vals.Get("scope")
 	} else {
 		var r struct {
-			AccessToken  string `json:"access_token"`
-			TokenType    string `json:"token_type"`
-			IDToken      string `json:"id_token"`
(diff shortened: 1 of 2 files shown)


ericchiang:
We'll that's terrible.

expires_in is defined as a JSON number[0], and JSON numbers aren't strings[1]. This is a busted implementation.

That being said, apparently paypall does the same thing[1] so I guess this is reasonable.

[0] https://tools.ietf.org/html/rfc6749#appendix-A.14
[1] https://tools.ietf.org/html/rfc4627#section-2.4
[2] https://github.com/golang/oauth2/blob/65a8d08c6292395d47053be10b3c5e91960def76/internal/token.go#L62

ericchiang:
Anyway, thanks for your contribution! In the future, please open an issue first then submit a PR. It ensures we can raise any issues before you go through the coding effort :)

PR #88, opened 2016-06-23
key: use shorter key ids

ericchiang:
Use shorter KeyIDs to reduce the overall size of JWTs produced by
this package.

The "kid" field value is unspecified and only exists to uniquely
identity keys in a JSONKeySet.[0] Therefore changing how we
generate this value shouldn't impact any existing programs.

This addresses coreos/dex#474

[0] https://tools.ietf.org/html/rfc7517#section-4.5

cc @chancez @sym3tri

### DIFF
--- a/key/key.go
+++ b/key/key.go
@@ -3,9 +3,9 @@ package key
 import (
 	"crypto/rand"
 	"crypto/rsa"
-	"encoding/base64"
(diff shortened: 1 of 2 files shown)


ericchiang:
Going to merge this unless anyone has concerns.

PR #90, opened 2016-07-07
Implement end_session_endpoint discovery metadata

ironcladlou:
Gather and expose end_session_endpoint metadata during discovery to
enable OP logout.

Spec: http://openid.net/specs/openid-connect-session-1_0.html#OPMetadata


ericchiang:
Sorry for the delayed reply, have been out of the office.

Should `check_session_iframe` be added too or is it not necessary for doing session management? I'm not totally familiar with this spec.

Also out of interest what providers implement this? I don't think I've run into one.

ironcladlou:
@ericchiang 
>Should check_session_iframe be added too or is it not necessary for doing session management? I'm not totally familiar with this spec.

According to the spec:

```
These OpenID Provider Metadata parameters MUST be included in the Server's discovery responses when Session Management and Discovery are supported:
```

So I suppose both should be supported in the client. I'm new to the spec and so am not yet sure whether `check_session_iframe` will be required for my use case. I'll see if I can get an answer to this.

>Also out of interest what providers implement this? I don't think I've run into one.

KeyCloak.

ericchiang:
Cool. Want to just add it for now then?

ironcladlou:
Sure! I'll add another commit for the additional field.

ironcladlou:
@ericchiang amended the commit to include both metadata fields and updated the commit log/PR description to match.

ericchiang:
lgtm. Thanks for your contribution!

PR #95, opened 2016-08-16
*: remove dependency on capnslog

ericchiang:
capnslog hijacks the standard loggers as an import side effect. For
some reason in dex this is causing the standard logger to not log
anything. For example net/http just swallows panics in handlers
without printing stack traces.

I'm having trouble tracking this down in dex, but removing the
capnslog's logging hijacker init() fixed this, so we're just going
to remove the dependency altogether. Importing go-oidc shouldn't
alter the standard logger anyway.


http/middleware.go
```diff
@@ -1,14 +0,0 @@
...
-)
-
-type LoggingMiddleware struct {
-	Next http.Handler
-}
-
-func (l *LoggingMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	log.Infof("HTTP %s %v", r.Method, r.URL)
-	l.Next.ServeHTTP(w, r)
-}
```
sym3tri: This is more than just a log implementation change. Why remove the middleware?
I'm ok with this if it wasn't being used (likely).

sym3tri:
otherwise LGTM

http/middleware.go
ericchiang: 1) We're not using it in dex.
2) This implementation is tied to the capnslog package so it seemed reasonable to remove as part of this diff. It's basically a "log to capnslog middleware"

ericchiang:
cool, merging

PR #97, opened 2016-08-18
oidc: verify claims before signatures

### LINKED ISSUE #93: IDTokenVerifier: Reject ID Tokens with invalid claims before verifying signature
This is an optimization brought up in kubernetes/kubernetes#30457.

There are certain claim values, such as an invalid issuer, that can be rejected before verifying the signature (a much heavier operation).

ericchiang:
Checking claims is a much lighter weight operation than verifying
signatures and re-syncing keys. While we still ensure that all JWTs
are correctly signed and valid, throw out tokens we know are invalid
based on claims before attempting to validate the signature.

Closes #93

### DIFF
--- a/oidc/verification.go
+++ b/oidc/verification.go
@@ -161,11 +161,18 @@ func NewJWTVerifier(issuer, clientID string, syncFunc func() error, keysFunc fun
 }
 
 func (v *JWTVerifier) Verify(jwt jose.JWT) error {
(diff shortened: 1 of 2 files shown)


ericchiang:
cc @chancez @sym3tri

chancez:
This seems reasonable to me.

Issue #98, opened 2016-08-27
Timing side channel in HMAC signature verification


kcq:
jose/sig_hmac.go:Verify doesn't use a constant time function to compare signatures and it introduces a timing side channel that can be used to forge signed tokens.

Should use hmac.Equal instead of bytes.Equal.

kcq:
That was pretty fast! Cool :) Hopefully you have a new Dex release coming up to take advantage of this fix.

ericchiang:
Thanks for the report! At some point we're going to switch dex to a more mature jose package (see #50) and hopefully have fewer of these kind of subtle issues. We'll also try to push a dex release soon with these changes.

ericchiang:
Actually, we've confirmed that this code isn't being used in dex so no need to do a release. We may remove this from the go-oidc package though.

PR #100, opened 2016-08-29
jose: remove hmac signing and verification code

ericchiang:
This was brought up during disucssions about #98. The HMAC logic
isn't used by this package or dex. We don't support client_secret_jwt
client authentication[0] in dex and don't provide an interface for it
in our client package.

Remove the HMAC code for now. We don't want to encourage its use and
can revert this change if ever need to support it.

This change has been tested with the dex repo and passes those unit
tests locally.

[0] https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

cc @sym3tri @gtank


sym3tri:
LGTM, I thought this was removed ages ago.

PR #101, opened 2016-08-29
oidc: implement new relaxed JWT verifier

pborzenkov:
Sometimes authenticator doesn't care which client the token was issued
to, but rather that the token is signed and valid overall. The 'aud'
claim is later checked by some other party that ensures that the client
actually exists and has rights to access the requested resource.

Right now it's required to create ad-hoc JWT verifiers to achieve this
behaviour, which implies fetching public key from OIDC provider on every
check.

To mitigate this, the patch adds new relaxed JWT verifier that follows
almost the same token validation procedure as the existing one except
for not validating the 'aud' claim.

Signed-off-by: Pavel Borzenkov pavel.borzenkov@gmail.com

### DIFF
--- a/oidc/verification.go
+++ b/oidc/verification.go
@@ -38,7 +38,11 @@ func containsString(needle string, haystack []string) bool {
 
 // Verify claims in accordance with OIDC spec
 // http://openid.net/specs/openid-connect-basic-1_0.html#IDTokenValidation
-func VerifyClaims(jwt jose.JWT, issuer, clientID string) error {
+func VerifyClaims(jwt jose.JWT, issuer, cliendID string) error {
+	return verifyClaims(jwt, issuer, cliendID, true)
+}
+
+func verifyClaims(jwt jose.JWT, issuer, clientID string, verifyAud bool) error {
 	now := time.Now().UTC()
 
 	claims, err := jwt.Claims()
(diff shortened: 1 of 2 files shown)


oidc/verification.go
```diff
@@ -142,29 +150,49 @@ func VerifyClientClaims(jwt jose.JWT, issuer string) (string, error) {
...
+		syncFunc:  syncFunc,
+		keysFunc:  keysFunc,
+		clock:     clockwork.NewRealClock(),
+		verifyAud: true,
+	}
+}
+
+// NewJWTVerifierRelaxed creates a verifier that follows OpenID Connect ID Token validation
+// procedure except for not validating the 'aud' claim.
+func NewJWTVerifierRelaxed(issuer string, syncFunc func() error, keysFunc func() []key.PublicKey) JWTVerifier {
```
ericchiang: Really dislike the name "NewJWTVerifierRelaxed" though yes we've seen this issue too.

pborzenkov: To be fair, I don't like it either. I had thought about the name for a while, and this one is the best I came up with. Maybe you have some suggestions?

You know, they say there are two hard things in software engineering: invalidating caches and naming things :)

ericchiang: > To be fair, I don't like it either. I had thought about the name for a while, and this one is the best I came up with. Maybe you have some suggestions?

I'll think about it, but the biggest part of any change like this is proposing the exact API changes. The naming, creating an API that might support additional validation tuning, etc. are things we should think about particularly since once it's in it's set in stone.

ericchiang:
Please look at the new verifier. It's more relaxed by default.

https://godoc.org/github.com/coreos/go-oidc#Provider.Verifier

PR #102, opened 2016-09-07
jose: use RSA PSS to Sign and Verify

satran:
Even though the standard points to PKCS1v15 as the recommended algorithm it has its flaws. Quoting the go documentation:

The original specification for encryption and signatures with RSA is PKCS#1 and the terms "RSA encryption" and "RSA signatures" by default refer to PKCS#1 version 1.5. However, that specification has flaws and new designs should use version two, usually called by just OAEP and PSS, where possible.

This commit changes the methods Verify and Sign to use PSS instead of PKCS1v15.

References:
- [Bleichenbacher's Paper: Chosen Ciphertext Attacks Against Protocols Based on the RSA Encryption Standard PKCS #1](http://archiv.infsec.ethz.ch/education/fs08/secsem/bleichenbacher98.pdf)
- [Coron et all. Paper: New Attacks on PKCS#1 v1.5 Encryption](https://www.iacr.org/archive/eurocrypt2000/1807/18070374-new.pdf)

### DIFF
--- a/jose/sig_rsa.go
+++ b/jose/sig_rsa.go
@@ -19,7 +19,7 @@ type SignerRSA struct {
 }
 
 func NewVerifierRSA(jwk JWK) (*VerifierRSA, error) {
-	if jwk.Alg != "" && jwk.Alg != "RS256" {
+	if jwk.Alg != "" && jwk.Alg != "PS256" {
 		return nil, fmt.Errorf("unsupported key algorithm %q", jwk.Alg)
 	}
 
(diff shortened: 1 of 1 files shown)


ericchiang:
We don't use RS256 because we feel it's a stronger algorithm, we use it because it's the default algorithm for OpenID Connect[0], and many providers don't let clients require a different one.

If you're using this jose library outside of OpenID Connect, I'd strongly recommend square's instead[1].

[0] https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
[1] https://github.com/square/go-jose

satran:
Fair enough. We were planning to use this library but thanks for pointing out alternatives.

PR #105, opened 2016-10-10
oauth2: don't send scopes to token endpoint

ericchiang:
Not specifying a "scope" parameter for a refresh token means using
the original scopes. However, it's currently impossible to redeem a
refresh token without specify these. This is very problematic when
a client doesn't know the original scopes.

Rework the parameters requested of the token endpoint to be more in
line with the spec. Also, never specifiy the scope parameter when
refreshing a token which is what golang.org/x/oauth2 does[0].

The behavior we care about is:
- Refresh tokens don't require "scope" and doesn't use "redirect_uri"[1]
- Access token requests don't use "scope"[2]

[0] https://github.com/golang/oauth2/blob/1e695b1c8febf17aad3bfa7bf0a819ef94b98ad5/oauth2.go#L230-L233
[1] https://tools.ietf.org/html/rfc6749#section-6
[2] https://tools.ietf.org/html/rfc6749#section-4.1.3

cc @sym3tri @dghubble

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -148,8 +148,11 @@ func (c *Client) HttpClient() phttp.Client {
 
 // Generate the url for initial redirect to oauth provider.
 func (c *Client) AuthCodeURL(state, accessType, prompt string) string {
-	v := c.commonURLValues()
+	v := url.Values{}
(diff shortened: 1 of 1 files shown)


ericchiang:
Done https://github.com/coreos/go-oidc/pull/110

PR #112, opened 2016-11-15
uses accesstoken instead of idtoken for refresh

### LINKED ISSUE #111: Refresh token requires id_token in response but openid spec it is optional
In the method [RefreshToken](https://github.com/coreos/go-oidc/blob/master/oidc/client.go#L755-L772), there is a line to parse the response from an OpenID provider.  The [current implementation expects](https://github.com/coreos/go-oidc/blob/master/oidc/client.go#L766) an `id_token`.  However in the [Openid spec](http://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse) it states that `id_token` is not required on the response.

jsloyer:
resolves #111

### DIFF
--- a/oidc/client.go
+++ b/oidc/client.go
@@ -763,7 +763,7 @@ func (c *Client) RefreshToken(refreshToken string) (jose.JWT, error) {
 		return jose.JWT{}, err
 	}
 
-	jwt, err := jose.ParseJWT(t.IDToken)
+	jwt, err := jose.ParseJWT(t.AccessToken)
 	if err != nil {
 		return jose.JWT{}, err
 	}


oidc/client.go
```diff
@@ -763,7 +763,7 @@ func (c *Client) RefreshToken(refreshToken string) (jose.JWT, error) {
 		return jose.JWT{}, err
 	}
 
-	jwt, err := jose.ParseJWT(t.IDToken)
+	jwt, err := jose.ParseJWT(t.AccessToken)
```
ericchiang: AccessTokens are opaque. Why would it be safe to assume they're a JWT?

jsloyer: @ericchiang I am fine leaving it with IDToken, since it seems most providers return that when `grant_type` is `refresh_token`.  I just need to go back to the OpenID provider Im looking at using and have them change their response as they don't return the `id_token`.

jsloyer: @ericchiang thoughts?

ericchiang: > AccessTokens are opaque. Why would it be safe to assume they're a JWT?
  
I don't see my initial question address.

I get what you're requesting: a flow for when an provider doesn't return an ID Token as part of the refresh response. I'm confused why this would be the solution.

PR #115, opened 2016-11-23
jose: use base64 raw URL encoding for modulus and exponent

pborzenkov:
Per JWS rfc7515 (Section 2. Terminology, Base64url Encoding), all
trailing '=' characters should be omitted. coreos/go-oidc/jose doesn't
implement this requirement.

While this was Ok when coreos/go-oidc/jose was used on both client and
server side, or when it was used on client side only (due to
decodeBase64URLPaddingOptional function), it stopped working when
go-oidc/jose is used on server side and stricter implementation of JWS
is used on the client side (notable, square/go-jose.v2 which is used by
new coreos/go-oidc package).

By following the RFC more strictly on the encoding side we:
  - allow to use new go-oidc package on the client side with
    DEX server v1, which relies on old go-oidc package;
  - don't break compatibility with existing software, as decoding
    handles both padded and unpadded messages.

### DIFF
--- a/jose/jwk.go
+++ b/jose/jwk.go
@@ -104,7 +104,7 @@ func encodeExponent(e int) string {
 			break
 		}
 	}
-	return base64.URLEncoding.EncodeToString(b[idx:])
+	return base64.RawURLEncoding.EncodeToString(b[idx:])
 }
 
 // Turns a URL encoded modulus of a key into a big int.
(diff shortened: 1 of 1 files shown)


ericchiang:
Nice catch. It looks like we're using `base64.URLEncoding` in a lot of places (this package predates `RawURLEncoding`), but there we correctly trim the `=`.

```
$ grep -r 'base64\.' *
http/http.go:	c, err := base64.StdEncoding.DecodeString(strings.TrimPrefix(auth, "Basic "))
jose/jose.go:	return base64.URLEncoding.DecodeString(seg)
jose/jose.go:	return strings.TrimRight(base64.URLEncoding.EncodeToString(seg), "=")
jose/jwk.go:	return base64.URLEncoding.EncodeToString(b[idx:])
jose/jwk.go:	return base64.URLEncoding.EncodeToString(n.Bytes())
jose/jwk.go:	return base64.URLEncoding.DecodeString(e)
oidc/util.go:	return fmt.Sprintf("%s@%s", base64.URLEncoding.EncodeToString(b), host), nil
verify.go:	payload, err := base64.RawURLEncoding.DecodeString(parts[1])
```

lgtm

PR #116, opened 2016-11-25
Use a token-aware HTTP client for UserInfo

jasonrig:
It seem that the current `UserInfo` function does not use the `tokenSource` to authenticate properly with the `/userinfo` endpoint (indeed `tokenSource` is not used at all in the `UserInfo` function. This PR replaces the `clientFromContext` call with an [`oauth2.NewClient`](https://godoc.org/golang.org/x/oauth2#NewClient) call that returns a token-aware HTTP client.

There are several other approaches that resolve non-authentication issue, such as:
* setting the client in the context and retaining the call to `clientFromContext`
* using the `oauth2.SetAuthHeader()` for the request that is passed to the HTTP client returned from `clientFromContext`

The latter option will always respect the HTTP client set by the context, but [the `SetAuthHeader` docs mention](https://github.com/golang/oauth2/blob/master/token.go#L78) that "... this method is unnecessary when using Transport or an HTTP Client returned by this package."

I'm not sure whether my proposed approach is optimal, and I'm happy to discuss.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -161,7 +161,19 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 	if p.userInfoURL == "" {
 		return nil, errors.New("oidc: user info endpoint is not supported by this provider")
 	}
-	resp, err := clientFromContext(ctx).Get(p.userInfoURL)
+
+	req, err := http.NewRequest("GET", p.userInfoURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("oidc: create GET request: %v", err)
+	}
+
+	token, err := tokenSource.Token()
+	if err != nil {
+		return nil, fmt.Errorf("oidc: get access token: %v", err)
+	}
+	token.SetAuthHeader(req)
+
+	resp, err := clientFromContext(ctx).Do(req)
 	if err != nil {
 		return nil, err
 	}


ericchiang:
Thanks for the thorough report.

Confirmed this worked by running the userinfo example app[0].

Looking into the oauth2 package, `oauth2.NewClient` actually grabs the transport for any custom client but not any other fields[1], so if someone has set a cookie jar or redirect check, `oauth2.NewClient` will override it.

A little petty, but for correctness maybe we can just use `SetAuthHeader` so other `http.Client` fields are preserved?

```go
req, err := http.NewRequest("GET", p.userInfoURL, nil)
if err != nil {
    return nil, fmt.Errorf("oidc: create GET request: %v", err)
}

token, err := tokenSource.Token()
if err != nil {
    return nil, fmt.Errorf("oidc: get access token: %v", err)
}
token.SetAuthHeader(req)

resp, err := clientFromContext(ctx).Do(req)
```

[0] https://github.com/coreos/go-oidc/blob/master/example/userinfo/app.go
[1] https://github.com/golang/oauth2/blob/d5040cddfc0da40b408c9a1da4728662435176a9/oauth2.go#L308

jasonrig:
Hey Eric, thanks for the positive feedback. You raise a good point and I'll
make the change shortly!

jasonrig:
@ericchiang I've updated the PR using the snippet you've provided. Tested locally and looks good to me :)

ericchiang:
lgtm. Can you squash the two commits into one?

https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

jasonrig:
rebased as requested.

ericchiang:
lgtm. thanks!

PR #117, opened 2016-11-28
jwks: fix panic when updating keys over HTTP2

pborzenkov:
Custom errContext implementation in jwks.go breaks the 'context'
contract that Err() should return a non-nil error after the context is
cancelled:

    // Err returns a non-nil error value after Done is closed.  Err returns
    // Canceled if the context was canceled or DeadlineExceeded if the
    // context's deadline passed.  No other values for Err are defined.
    // After Done is closed, successive calls to Err return the same value.
    Err() error

This causes the following panic in http2 code:

```
panic: err must be non-nil

goroutine 154 [running]:
panic(0x9ac2a0, 0xc42020ff50)
        /usr/lib/go/src/runtime/panic.go:500 +0x1a1
net/http.(*http2pipe).closeWithError(0xc420098de8, 0xc420098e38, 0x0, 0x0, 0x0)
        /usr/lib/go/src/net/http/h2_bundle.go:2775 +0x20b
net/http.(*http2pipe).CloseWithError(0xc420098de8, 0x0, 0x0)
        /usr/lib/go/src/net/http/h2_bundle.go:2762 +0x53
net/http.(*http2clientStream).awaitRequestCancel(0xc420098dc0, 0xc420110f00)
        /usr/lib/go/src/net/http/h2_bundle.go:5074 +0x1e9
created by net/http.(*http2clientConnReadLoop).handleResponse
        /usr/lib/go/src/net/http/h2_bundle.go:6189 +0x63b
```

Fix this by using golang.org/x/sync/singleflight package, which allows
to suppress duplicate function execution. As a bonus, we no longer need
a goroutine here and don't need to recover from panic.

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -39,38 +39,39 @@ type remoteKeySet struct {
 	// guard all other fields
 	mu sync.Mutex
 
-	// inflightCtx is the context of the current HTTP request to update the keys.
-	// Its Err() method returns any errors encountered during that attempt.
+	// inflightCtx suppresses parallel execution of updateKeys and allows
+	// multiple goroutines to wait for its result.
+	// Its Err() method returns any errors encountered during updateKeys.
 	//
-	// If nil, there is no inflight request.
-	inflightCtx context.Context
+	// If nil, there is no inflight updateKeys request.
+	inflightCtx *inflight
 
 	// A set of cached keys and their expiry.
 	cachedKeys []jose.JSONWebKey
 	expiry     time.Time
 }
 
-// errContext is a context with a customizable Err() return value.
-type errContext struct {
-	context.Context
-
-	cf  context.CancelFunc
-	err error
+// inflight is used to wait on some in-flight request from multiple goroutines
+type inflight struct {
+	done chan struct{}
+	err  error
 }
 
-func newErrContext(parent context.Context) *errContext {
-	ctx, cancel := context.WithCancel(parent)
-	return &errContext{ctx, cancel, nil}
+// Done returns a channel that is closed when the inflight request finishes.
+func (i *inflight) Done() <-chan struct{} {
+	return i.done
 }
 
-func (e errContext) Err() error {
-	return e.err
+// Err returns any error encountered during request execution. May be nil.
+func (i *inflight) Err() error {
+	return i.err
 }
 
-// cancel cancels the errContext causing listeners on Done() to return.
-func (e errContext) cancel(err error) {
-	e.err = err
-	e.cf()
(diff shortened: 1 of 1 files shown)


ericchiang:
Nice! Apparently non of my test apps are using HTTP/2 :/

Fix largely lgtm, though I think that goroutine you mentioned is now out of our reach[0].

I generally dislike adding more dependencies, but I've tried to write this particularly piece of code and it's clearly pretty subtle.

I think a solution without singleflight is that `inlightCtx` shouldn't be a Context but instead just a channel with a wait condition. Have it do it's same signaling, but use `r.ctx` instead of `inflightCtx` here[1]. `inflightCtx` then becomes something like this:

```go
type inflight struct {
	done chan struct{}
	err  error
}

func (i *inflight) Done() <-chan struct{} { return i.done }
func (i *inflight) Err() error            { return i.err }
func (i *inflight) Cancel(e error)        { i.err = e; close(i.done) }
```

Any opinions on why we should add a new dependency vs. just cleaning up the existing code?

[0] https://github.com/golang/sync/blob/1ae7c7b/singleflight/singleflight.go#L86
[1] https://github.com/coreos/go-oidc/blob/d3e23e1/jwks.go#L134

pborzenkov:
> though I think that goroutine you mentioned is now out of our reach[0].

No, I use plan group.Do() which does the function call in the same goroutine.
DoChan() does indeed spawns news goroutine, but we don't need to use it here.
Simple Do() would suffice.

ericchiang:
> No, I use plain group.Do() which does the function call in the same goroutine.

Oh, that's a problem though. Updating keys is currently tied to the global context, not the individual request's context. This is because individual requests may cancel their context, but the global update of keys shouldn't.

Also, the context is now only passed if there are no other concurrent callers. Consider the following scenario:

1. Caller A triggers a key update.
2. Caller B blocks but its context isn't passed.
3. Caller B cancel's it's context.

Caller B's cancel isn't actually propagated because B is now tied to  A's context, not its own.

I know this sounds pedantic, but this come up a lot when the provider is unavailable. An edge case we hit a lot in Kubernetes.

pborzenkov:
> Caller B's cancel isn't actually propagated because B is now tied to A's context, not its own.

Yeah, haven't thought about such scenario. In this case, we either need to switch to DoChan() (and lose control over goroutine), or just rework inflightCtx not to behave like a Context, but rather just signal an error to parent goroutine.

I'll go with the second choice and rework the patch soon.

ericchiang:
lgtm. Thanks!

PR #118, opened 2016-11-28
key: fix test fail introduced by e7893b34

pborzenkov:
Commit e7893b34 changed the way keys' exponent and modulus are encoded,
but didn't fix the test to expect new format.

Fix the test here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ go:
   - 1.6.3
 
 install:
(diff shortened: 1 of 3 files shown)


ericchiang:
Ah, travis isn't testing subpackages anymore. As part of this PR, could turn that back on?

ericchiang:
Thanks! lgtm.

PR #122, opened 2016-12-30
Skip checking issuer when creating Provider.

### LINKED ISSUE #121: Azure AD v2.0 fails issuer check when creating provider
[Azure AD v2.0](https://docs.microsoft.com/en-us/azure/active-directory/active-directory-v2-tokens) provides OIDC config via:
https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration

Issuer in the config json is returned as
`https://login.microsoftonline.com/{tenantid}/v2.0`

As a result `NewProvider()` method fails.

TamalSaha:
Fixes #121

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


TamalSaha:
@ericchiang Here is an example using Azure AD v2.0 https://github.com/appscode/gologin/blob/master/azure/login.go#L20

ericchiang:
Per conversation in #121 I don't think we want to provide the ability to skip the issuer comparison. For example, in #121 this check was actually a good indication that the wrong issuer ID was being provided.

Okay for me to close?

TamalSaha:
@ericchiang I think using `9188040d-6c67-4c5b-b112-36a304b66dad` instead of `common` results in proper issuer check. Closing this issue.

Issue #123, opened 2017-01-04
Verify does not check if ID Token has expired or not


xeonx:
The VerifyExpiry() option as no impact on the result of the Verify method

rithujohn191:
Thank you for your comment. You are right, the VerifyExpiry() only updates the verificationConfig. The Verify function does not check to ensure that the current time is before the token Expiry time . We will rectify this.

PR #124, opened 2017-01-06
*: check if ID token has expired.

rithujohn191:
Addresses #123.

Also added a unit test to verify_test.go

@sym3tri

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -145,6 +145,13 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		}
 	}
 
+	// If a checkExpiry is specified, make sure token is not expired.
+	if v.config.checkExpiry != nil {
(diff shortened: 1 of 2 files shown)


verify.go
```diff
@@ -145,6 +145,13 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		}
 	}
 
+	// If a checkExpiry is specified, make sure token is not expired.
+	if v.config.checkExpiry != nil {
```
kans: Would this ever be anything other than time.Now?

squat: you might want to preemptively check if it will expire next week

rithujohn191: Looks like VerifyExpiry() always sets "checkExpiry = time.Now"

rithujohn191: @squat Verifying token expiry is optional, it is implemented only if checkExpiry is set by VerifyExpiry().

PR #129, opened 2017-01-18
*: Update ID token Validation according to OIDC spec.

### LINKED ISSUE #128: ID Token Validation must check expiry and client_id by default.
According to (http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) the client must validate that the audience Claim contains the client_id value and check expiry by default. Currently they are optional in the Verify function.

rithujohn191:
Closes #128.

The current ID Token validation is driven by the VerificationOptions argument provided to the Verifier. Thus we do not implement some of "Must Checks" in the OIDC spec (https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation).

This change ensures that the ID Token validation process has stronger defaults, in turn making it more secure. We still provide the option to skip these default checks via the configuration attributes in the Config struct.

As a part of this change I also updated the example apps and the tests in verify_test.go.

### DIFF
--- a/example/idtoken/app.go
+++ b/example/idtoken/app.go
@@ -27,7 +27,11 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	verifier := provider.Verifier()
+	oidcConfig := &oidc.Config{
+		ClientID:       clientID,
+		SkipNonceCheck: true,
+	}
+	verifier := provider.Verifier(oidcConfig)
 
 	config := oauth2.Config{
 		ClientID:     clientID,
--- a/example/nonce/app.go
+++ b/example/nonce/app.go
@@ -23,10 +23,7 @@ var (
 
 const appNonce = "a super secret nonce"
 
-// Create a nonce source.
-type nonceSource struct{}
-
-func (n nonceSource) ClaimNonce(nonce string) error {
+func ClaimNonce(nonce string) error {
 	if nonce != appNonce {
 		return errors.New("unregonized nonce")
 	}
@@ -41,8 +38,12 @@ func main() {
 		log.Fatal(err)
 	}
 
+	oidcConfig := &oidc.Config{
+		ClientID:   clientID,
+		ClaimNonce: ClaimNonce,
+	}
 	// Use the nonce source to create a custom ID Token verifier.
-	nonceEnabledVerifier := provider.Verifier(oidc.VerifyNonce(nonceSource{}))
+	nonceEnabledVerifier := provider.Verifier(oidcConfig)
 
 	config := oauth2.Config{
 		ClientID:     clientID,
--- a/verify.go
+++ b/verify.go
@@ -17,53 +17,59 @@ import (
 // IDTokenVerifier provides verification for ID Tokens.
 type IDTokenVerifier struct {
 	keySet *remoteKeySet
-	config *verificationConfig
-}
-
-// verificationConfig is the unexported configuration for an IDTokenVerifier.
-//
-// Users interact with this struct using a VerificationOption.
-type verificationConfig struct {
+	config *Config
 	issuer string
-	// If provided, this value must be in the ID Token audiences.
-	audience string
-	// If not nil, check the expiry of the id token.
-	checkExpiry func() time.Time
-	// If specified, only these sets of algorithms may be used to sign the JWT.
-	requiredAlgs []string
-	// If not nil, don't verify nonce.
-	nonceSource NonceSource
 }
 
-// VerificationOption provides additional checks on ID Tokens.
-type VerificationOption interface {
-	// Unexport this method so other packages can't implement this interface.
-	updateConfig(c *verificationConfig)
+// Config is the configuration for an IDTokenVerifier.
+type Config struct {
+	// Expected audience of the token. For a majority of the cases this is expected to be
+	// the ID of the client that initialized the login flow. It may occasionally differ if
+	// the provider supports the authorizing party (azp) claim.
+	//
+	// If not provided, users must explicitly set SkipClientIDCheck.
+	ClientID string
+	// Method to verify the ID Token nonce. If a nonce is present and this method
+	// is nil, users must explicitly set SkipNonceCheck.
+	//
+	// If the ID Token nonce is empty, for example if the client didn't provide a nonce in
+	// the initial redirect, this may be nil.
+	ClaimNonce func(nonce string) error
+	// If specified, only this set of algorithms may be used to sign the JWT.
+	//
+	// Since many providers only support RS256, SupportedSigningAlgs defaults to this value.
+	SupportedSigningAlgs []string
+
+	// If true, no ClientID check performed. Must be true if ClientID field is empty.
+	SkipClientIDCheck bool
+	// If true, token expiry is not checked.
+	SkipExpiryCheck bool
+	// If true, nonce claim is not checked. Must be true if ClaimNonce field is empty.
+	SkipNonceCheck bool
+
+	// Time function to check Token expiry. Defaults to time.Now
+	Now func() time.Time
 }
 
 // Verifier returns an IDTokenVerifier that uses the provider's key set to verify JWTs.
 //
 // The returned IDTokenVerifier is tied to the Provider's context and its behavior is
 // undefined once the Provider's context is canceled.
-func (p *Provider) Verifier(options ...VerificationOption) *IDTokenVerifier {
-	config := &verificationConfig{issuer: p.issuer}
-	for _, option := range options {
-		option.updateConfig(config)
-	}
+func (p *Provider) Verifier(config *Config) *IDTokenVerifier {
 
-	return newVerifier(p.remoteKeySet, config)
+	return newVerifier(p.remoteKeySet, config, p.issuer)
 }
 
-func newVerifier(keySet *remoteKeySet, config *verificationConfig) *IDTokenVerifier {
-	// As discussed in the godocs for VerifrySigningAlg, because almost all providers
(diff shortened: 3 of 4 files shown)


verify.go
```diff
@@ -17,53 +17,51 @@ import (
...
-	// If not nil, check the expiry of the id token.
-	checkExpiry func() time.Time
+// Config is the configuration for an IDTokenVerifier.
+type Config struct {
+	// If provided, ClientID must be in the ID Token audiences.
+	// It ensures that an ID Token was issued for the specific client.
+	ClientID string
+	// If provided, verify nonce.
+	ClaimNonce func(nonce string) error
 	// If specified, only these sets of algorithms may be used to sign the JWT.
```
ericchiang: Just noticed a typo. s/only these sets/only this set/g

verify.go
```diff
@@ -192,79 +199,22 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
-		if err := v.config.nonceSource.ClaimNonce(t.Nonce); err != nil {
-			return nil, err
+	// If SkipNonceCheck is not set ClaimNonce cannot be Nil.
+	if !(v.config.SkipNonceCheck) {
+		if v.config.ClaimNonce != nil {
+			if err := v.config.ClaimNonce(t.Nonce); err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, fmt.Errorf("oidc: Invalid configuration. ClaimNonce must be provided or SkipNonceCheck must be set.")
```
ericchiang: If there's no nonce present, you shouldn't have to specify a `ClaimNonce` function.

verify.go
```diff
@@ -17,53 +17,51 @@ import (
...
-	requiredAlgs []string
-	// If not nil, don't verify nonce.
-	nonceSource NonceSource
-}
+	// Defaults to RS256.
+	SupportedSigningAlgs []string
+
+	// If true, no ClientID check performed. Must be true if ClientID field is empty.
+	SkipClientIDCheck bool
+	// If true, Token expiry is not checked.
```
ericchiang: "Token" and "Nonce" should be lower cased.

verify.go
```diff
@@ -17,53 +17,51 @@ import (
...
+	// It ensures that an ID Token was issued for the specific client.
+	ClientID string
+	// If provided, verify nonce.
+	ClaimNonce func(nonce string) error
 	// If specified, only these sets of algorithms may be used to sign the JWT.
-	requiredAlgs []string
-	// If not nil, don't verify nonce.
-	nonceSource NonceSource
-}
+	// Defaults to RS256.
```
ericchiang: Can we expand this doc to include some of the existing signing algorithm comment? https://godoc.org/github.com/coreos/go-oidc#VerifySigningAlg

verify.go
```diff
@@ -17,53 +17,51 @@ import (
...
-	issuer string
-	// If provided, this value must be in the ID Token audiences.
-	audience string
-	// If not nil, check the expiry of the id token.
-	checkExpiry func() time.Time
+// Config is the configuration for an IDTokenVerifier.
+type Config struct {
+	// If provided, ClientID must be in the ID Token audiences.
+	// It ensures that an ID Token was issued for the specific client.
+	ClientID string
```
ericchiang: So `ClientID` might need to match either `aud` or `azp`: https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

Do we want to do anything special here? Like say that if there are multiple audiences `ClientID` should match the authorizing party field? Is that too magic?

verify.go
```diff
@@ -17,53 +17,51 @@ import (
...
-// Users interact with this struct using a VerificationOption.
-type verificationConfig struct {
-	issuer string
-	// If provided, this value must be in the ID Token audiences.
-	audience string
-	// If not nil, check the expiry of the id token.
-	checkExpiry func() time.Time
+// Config is the configuration for an IDTokenVerifier.
+type Config struct {
+	// If provided, ClientID must be in the ID Token audiences.
```
ericchiang: This comment seems to imply that it's okay to not provide this value. We might want to change it to say something like.

```
// The ID of the client that started the login flow, which is expected to be in the ID Token's audiences.
//
// If not provided, users must explicitly set SkipClientIDCheck.
```

verify.go
```diff
@@ -134,35 +132,44 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
+	if t.Issuer != v.issuer {
+		return nil, fmt.Errorf("oidc: id token issued by a different provider, expected %q got %q", v.issuer, t.Issuer)
 	}
 
-	// If a client ID has been provided, make sure it's part of the audience.
-	if v.config.audience != "" {
-		if !contains(t.Audience, v.config.audience) {
-			return nil, fmt.Errorf("oidc: expected audience %q got %q", v.config.audience, t.Audience)
+	// If a client ID has been provided, make sure it's part of the audience. SkipClientIDCheck must be true if ClientID is empty.
+	if !(v.config.SkipClientIDCheck) {
```
ericchiang: These parentheses aren't needed.

ericchiang:
Some comments about why we're making this change (having stronger defaults) would be nice too.

verify.go
rithujohn191: So setting SkipNonceCheck to true is not mandatory if ClaimNonce is not provided right? Or would you like me to enforce this by adding another condition of IDToken.Nonce == nil?

ericchiang: 1. Nonce is returned in id token claims. SkipNonceCheck must be true or ClaimNonce must not be nil.
2. No nonce is returne in id token claims. Don't require either.

rithujohn191:
The current ID Token validation is driven by the VerificationOptions argument provided to the Verifier. Thus we do not implement some of "Must Checks" in the OIDC spec (https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation). 

This change ensures that the ID Token validation process has stronger defaults, in turn making it more secure. We still provide the option to skip these default checks via the configuration attributes in the Config struct.

verify.go
rithujohn191: Added a check which sets SkipNonceCheck == true if no nonce is returned in the id token claims.

ericchiang:
@rithujohn191 in the commit message :)

example/nonce/app.go
```diff
@@ -41,8 +38,12 @@ func main() {
 		log.Fatal(err)
 	}
 
+	oidc_config := &oidc.Config{
```
ericchiang: nit: use camelCase, not snake_case.

verify.go
ericchiang: An individual call to Verify shouldn't mutate the state of the IDTokenVerifier

verify.go
```diff
@@ -17,53 +17,53 @@ import (
...
-	nonceSource NonceSource
 }
 
-// VerificationOption provides additional checks on ID Tokens.
-type VerificationOption interface {
-	// Unexport this method so other packages can't implement this interface.
-	updateConfig(c *verificationConfig)
+// Config is the configuration for an IDTokenVerifier.
+type Config struct {
+	// The ID of the client that started the login flow, which is expected to be in the ID Token's audiences.
```
ericchiang: > The ID of the client that started the login flow,

This isn't always true. We probably want something like

> Expected audience of the token. For a majority of the cases this is expected to be the ID of
> the client that initialized the login flow. It may occasionally differ if the provider supports the
> authorizing party (azp) claim.
>
> If not provided, users must explicitly set SkipClientIDCheck.

example/idtoken/app.go
```diff
@@ -27,7 +27,11 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	verifier := provider.Verifier()
+	oidc_config := &oidc.Config{
```
ericchiang: same nit: use camelCase here.

verify.go
```diff
@@ -133,36 +133,51 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		claims:   payload,
 	}
 
-	// Check issuer.
-	if t.Issuer != v.config.issuer {
-		return nil, fmt.Errorf("oidc: id token issued by a different provider, expected %q got %q", v.config.issuer, t.Issuer)
+	if t.Nonce == nil {
```
ericchiang: can this be nil? I thought it was a string

verify.go
rithujohn191: Would you like me to handle it via if statements then? 

IDToken.Nonce == "" then we would want to skip the check all together right? Even if SkipNonceCheck == true. Maybe we should handle that corner case too.

ericchiang: something like

```
if IDToken.Nonce != "" && !SkipNonceCheck {
    // check nonce
}
```

verify.go
```diff
@@ -17,53 +17,55 @@ import (
...
-	updateConfig(c *verificationConfig)
+// Config is the configuration for an IDTokenVerifier.
+type Config struct {
+	// Expected audience of the token. For a majority of the cases this is expected to be
+	// the ID of the client that initialized the login flow. It may occasionally differ if
+	// the provider supports the authorizing party (azp) claim.
+	//
+	// If not provided, users must explicitly set SkipClientIDCheck.
+	ClientID string
+	// If provided, verify nonce.
```
ericchiang: Maybe something like this?

> Method to verify the ID Token nonce. If a nonce is present and this method
> is nil, users must explicitly set SkipNonceCheck.
>
> If the ID Token nonce is empty, (for example if it's unused), this may be nil.

PR #130, opened 2017-01-20
Redirection URL

gambol99:
- exposing the redirection url post client creation; the particular use case being able to change this setting on the fly (such as Host header)

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -141,6 +141,22 @@ func NewClient(hc phttp.Client, cfg Config) (c *Client, err error) {
 	return
 }
 
+// Return the redirection url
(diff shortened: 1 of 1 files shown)


ericchiang:
This is a data race, you'd need to protect redirectURL with a mutex to do this.

I'm generally against getters an setters like this. It's easy enough to do

```
func newClient(redirectURL string) *oauth2.Client
```

then call that when you need a new client. We shouldn't be opinionated about what subset of fields in the client might need to be dynamically changed.

In addition we generally think of this package as deprecated. Users should migrate to `golang.org/x/oauth2` as it's much better maintained and correct. We'd like to only start accepting bug fixes for these packages.

PR #131, opened 2017-01-26
add an option to allow the issuer claim to be schemaless

### LINKED ISSUE #125: Issuer mismatch for Google: accounts.google.com vs https://accounts.google.com
According to [Google's Auth documentation](https://developers.google.com/identity/sign-in/web/backend-auth), a token will contain an issuer of `accounts.google.com` **or** `https://accounts.google.com`. In my testing, when creating a provider with `NewProvider`, the `https://...` is required, but decoded raw auth tokens contain just `accounts.google.com`, which causes an [issuer mismatch and failed verification](https://github.com/coreos/go-oidc/blob/master/oidc.go#L105). [My solution](https://github.com/bcspragu/go-oidc/commit/1ccfcb33c8965bf5dea617a31bd7919cc1be2759) was to explicitly add new issuers to `IDTokenVerifier` with an `AddIssuer` method, but substring matching or ignoring the protocol would also work, but is arguably less secure.

ericchiang:
This is because Google often returns "account.google.com" instead
of the spec compliant "https://account.google.com".

Fixes #125

cc @rithujohn191 don't know if I love the name of the added field. Any opinions?

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -47,6 +47,12 @@ type Config struct {
 	// If true, nonce claim is not checked. Must be true if ClaimNonce field is empty.
 	SkipNonceCheck bool
 
+	// Google sometimes return "accounts.google.com" as the ID Token issuer, instead of
+	// the spec required value "https://accounts.google.com".
+	//
+	// Enable this option to tolerate when the issuer claim is schemaless.
+	SkipIssuerSchemaCheck bool
+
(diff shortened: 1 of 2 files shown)


ksswr:
It is scheme, not schema

ericchiang:
Blah, thanks @ksswr! Let me fix that.

PR #132, opened 2017-01-27
special case google accounts issuer validation

### LINKED ISSUE #125: Issuer mismatch for Google: accounts.google.com vs https://accounts.google.com

ericchiang:
Google sometimes returns "accounts.google.com" instead of
"https://accounts.google.com". Special case this.

Originally we merged a new verification flag, but since this should
only be used with Google, and should always be on with Google, it
makes more sense for us to detect this case transparently.

Fixes #125

cc @ksswr @rithujohn191

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -201,19 +201,35 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 // The ID Token only holds fields OpenID Connect requires. To access additional
 // claims returned by the server, use the Claims method.
 type IDToken struct {
-	// The URL of the server which issued this token. This will always be the same
-	// as the URL used for initial discovery.
+	// The URL of the server which issued this token. OpenID Connect
+	// requires this value always be identical to the URL used for
+	// initial discovery.
+	//
+	// Note: Because of a known issue with Google Accounts' implementation
+	// this value may differ when using Google.
+	//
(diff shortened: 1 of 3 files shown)


ericchiang:
@rithujohn191 can you take another look?

rithujohn191:
Maybe we could add a documentation note somewhere clarifying the behavior we expect for issuer schemes?

ericchiang:
@rithujohn191 I can expand the internal docs but I don't think we should advertise this in the package docs.

ericchiang:
more docs added

rithujohn191:
lgtm

PR #135, opened 2017-02-27
*: use std library's context package

ericchiang:
cc @rithujohn191


ericchiang:
@rithujohn191 something was flaky in the old "keys" package. Can't reproduce because I'm using Go1.7 on my machine. Anyway, all green now.

PR #138, opened 2017-03-07
oidc: allow `none` in TokenEndpointAuthMethodsSupported

### LINKED ISSUE #137: Error when token_endpoint_auth_methods_supported contains none
I'm attempting to use this library via Kubernetes with [okta](https://www.okta.com/) as an IDP.

When calling [oidc.FetchProviderConfig](https://github.com/kubernetes/kubernetes/blob/v1.5.3/plugin/pkg/auth/authenticator/token/oidc/oidc.go#L176) validation failes giving the error `token_endpoint_auth_signing_alg_values_supported cannot include 'none'` which comes from [this line](https://github.com/coreos/go-oidc/blob/master/oidc/provider.go#L357)

Okta's `/.well-known/openid-configuration` endpoint returns the following 
```json
...
"token_endpoint_auth_methods_supported": [
        "client_secret_basic",
        "client_secret_post",
        "none"
    ]
```
After reading relevant portions of the OIDC discovery spec I don't see why go-oidc fails if `none` is in `token_endpoint_auth_methods_supported`

From https://openid.net/specs/openid-connect-discovery-1_0.html#rfc.section.3

> token_endpoint_auth_methods_supported
 > OPTIONAL. JSON array containing a list of Client Authentication methods supported by this Token Endpoint. The options are client_secret_post, client_secret_basic, client_secret_jwt, and private_key_jwt, as described in Section 9 of OpenID Connect Core 1.0 [OpenID.Core]. Other authentication methods MAY be defined by extensions. If omitted, the default is client_secret_basic -- the HTTP Basic Authentication Scheme specified in Section 2.3.1 of OAuth 2.0 [RFC6749].

It seems like `none` should be an allowed entry in `token_endpoint_auth_methods_supported` am I completely off base here?

curtisallen:
Maintenance patch for the deprecated `oidc` package
Fixes #137

cc:/ @ericchiang

### DIFF
--- a/oidc/provider.go
+++ b/oidc/provider.go
@@ -353,9 +353,6 @@ func (p ProviderConfig) Valid() error {
 	if !contains(p.IDTokenSigningAlgValues, "RS256") {
 		return errors.New("id_token_signing_alg_values_supported must include 'RS256'")
 	}
-	if contains(p.TokenEndpointAuthMethodsSupported, "none") {
-		return errors.New("token_endpoint_auth_signing_alg_values_supported cannot include 'none'")
-	}
 
 	uris := []struct {
 		val      *url.URL
--- a/oidc/provider_test.go
+++ b/oidc/provider_test.go
@@ -183,6 +183,40 @@ func TestProviderConfigUnmarshal(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			// token_endpoint_auth_methods_supported can contain `none`
+			data: `{
+				"issuer": "https://server.example.com",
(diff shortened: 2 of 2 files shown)


ericchiang:
lgtm

curtisallen:
Thanks for the speady review and merge Eric! 🙇

ericchiang:
@curtisallen do you want to open the upstream PR to bump this dependency too? I can if you're busy.

curtisallen:
@ericchiang yup! Doing so right now :D

Issue #140, opened 2017-04-14
Failure on missing key id


Grillz:
Hey, I'm trying to use this library to connect to https://github.com/ory/hydra. Everything works fine up until the final signature verification at https://github.com/coreos/go-oidc/blob/master/verify.go#L185 . The JWT headers that Hydra provides do not have `kid` as a part of them. Here is a comparison:

Hydra   
```
{
  "alg": "RS256",
  "typ": "JWT"
}
```
Google   
```
{
  "alg": "RS256",
  "kid": "fb41c2409340c77eb259a66c9f79809649a62971"
}
```

Note, the RFC on JWT's doesn't specify a `kid` parameter https://tools.ietf.org/html/rfc7519#section-5

When the verification step happens it looks for the `KeyId` that was parsed by go-jose, because its nil it fails https://github.com/coreos/go-oidc/blob/master/verify.go#L193.

I'd be happy to rewrite this algorithm 👍

ericchiang:
Sounds good. It sounds like if there aren't any key IDs provided in the JWT, then we should check the signature against all the keys. Feel free to send a patch!

Grillz:
exactly, I'll get started on this shortly!

ericchiang:
Actually looking into this it seems like this library should work fine if the JWT doesn't specify a key ID and the JWK's don't have a key ID. They'll both be the empty string and everything should work.

The condition you're hitting implies that JWT isn't signed with a signing algorithm that we allow by default. What's the actual error that method is returning?

Grillz:
Doesn't https://github.com/coreos/go-oidc/blob/master/verify.go#L193 break it? If there are no key ids present it equals 0, currently its appending a blank space `" "`, and then failing when it can't find that key id on https://github.com/coreos/go-oidc/blob/master/verify.go#L198

ericchiang:
> If there are no key ids present it equals 0

Not quite. That condition only happens if there are no signatures with an allows signing algorithm.
```go
	var keyIDs, gotAlgs []string
	for _, sig := range jws.Signatures {
		if len(v.config.SupportedSigningAlgs) == 0 || contains(v.config.SupportedSigningAlgs, sig.Header.Algorithm) {
			keyIDs = append(keyIDs, sig.Header.KeyID)
		} else {
			gotAlgs = append(gotAlgs, sig.Header.Algorithm)
		}
	}
	if len(keyIDs) == 0 {
[...]

PR #141, opened 2017-04-18
Support missing key id

Grillz:
Per https://github.com/coreos/go-oidc/issues/140, I've implemented a new algorithm in the `verify` function, that first looks for key ids, if they are not present then searches by algorithm.

### DIFF
--- a/jose_test.go
+++ b/jose_test.go
@@ -277,13 +277,29 @@ var (
 
 	testKeyRSA_2048_0 = mustLoadJWK(`{
 		"kty": "RSA",
+		"alg": "RS256",
 		"kid": "13b668c7b4f5d46d53de65e4b9f9c522fad4870e4d1656e459c88de6e90984d6",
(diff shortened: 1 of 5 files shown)


jwks.go
```diff
@@ -156,6 +156,89 @@ func (r *remoteKeySet) keysWithID(ctx context.Context, keyIDs []string) ([]jose.
...
+}
+func (r *remoteKeySet) keysWithAlgs(ctx context.Context, algs []string) ([]jose.JSONWebKey, error) {
+	keys, ok := r.keysWithAlgsFromCache(algs)
+	if ok {
+		return keys, nil
+	}
+
+	var inflightCtx *inflight
+	func() {
+		r.mu.Lock()
```
ericchiang: I really dislike replicating this logic. It's fragile because it's dealing with the retry/sync mechanism.

jose_test.go
```diff
@@ -277,10 +277,17 @@ var (
 
 	testKeyRSA_2048_0 = mustLoadJWK(`{
 		"kty": "RSA",
+		"alg": "RS256",
```
ericchiang: This is a generated file, but I don't see a change to the generator.

Grillz:
Closing per https://github.com/coreos/go-oidc/issues/140

PR #145, opened 2017-04-26
*: various cleanups and improved unit testing

ericchiang:
Biggest cleanup is that the remoteKeySet now does the key ID matching
instead of the idTokenVerifier.

On the testing side, added tests that actually exercise the expected
caching behavior.


ericchiang:
Still need to test this against Google, Dex and Okta.

ericchiang:
@rithujohn191 as a follow up I'd like to create a dex equivalent of the example-app and add some docs around testing. Mind if I hold off on that until this gets reviewed?

ericchiang:
bump @rithujohn191

jwks.go
```diff
@@ -38,147 +39,157 @@ type remoteKeySet struct {
...
+// result cannot be called until the wait() channel has returned a value.
+func (i *inflight) result() ([]jose.JSONWebKey, error) {
+	return i.keys, i.err
+}
 
-	// Have the keys expired?
-	if expiry.Add(keysExpiryDelta).Before(r.now()) {
-		return nil, false
+func (r *remoteKeySet) verify(ctx context.Context, jws *jose.JSONWebSignature) ([]byte, error) {
+	// We don't support JWTs signed with multiple signatures.
```
ericchiang: We've never supported multiple keys and the user will actually get a super explicit error message if they try now:

```
oidc: multiple signatures on id token not supported
```

jwks.go
```diff
@@ -38,147 +39,157 @@ type remoteKeySet struct {
...
+		// once the goroutine is done.
+		go func() {
+			// Sync keys and finish inflight when that's done.
+			keys, expiry, err := r.updateKeys()
+
+			r.inflight.done(keys, err)
+
+			// Lock to update the keys and indicate that there is no longer an
+			// inflight request.
+			r.mu.Lock()
```
ericchiang: we're in a goroutine here :)

PR #148, opened 2017-05-12
Check response content-type to improve message if cannot decode as JSON

chancez:
If the Content-Type is not "application/json", add extra text indicating
that the response was not JSON before propagating the unmarshal error to
the caller.

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -2,7 +2,6 @@ package oidc
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -182,14 +181,16 @@ func (r *remoteKeySet) updateKeys() ([]jose.JSONWebKey, time.Time, error) {
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return nil, time.Time{}, fmt.Errorf("oidc: read response body: %v", err)
+		return nil, time.Time{}, fmt.Errorf("unable to read response body: %v", err)
 	}
+
 	if resp.StatusCode != http.StatusOK {
 		return nil, time.Time{}, fmt.Errorf("oidc: get keys failed: %s %s", resp.Status, body)
 	}
 
 	var keySet jose.JSONWebKeySet
-	if err := json.Unmarshal(body, &keySet); err != nil {
+	err = unmarshalResp(resp, body, &keySet)
+	if err != nil {
 		return nil, time.Time{}, fmt.Errorf("oidc: failed to decode keys: %v %s", err, body)
 	}
 
--- a/oidc.go
+++ b/oidc.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"mime"
 	"net/http"
 	"strings"
 	"time"
@@ -93,18 +94,23 @@ func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to read response body: %v", err)
 	}
+
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("%s: %s", resp.Status, body)
 	}
-	defer resp.Body.Close()
+
 	var p providerJSON
-	if err := json.Unmarshal(body, &p); err != nil {
+	err = unmarshalResp(resp, body, &p)
+	if err != nil {
 		return nil, fmt.Errorf("oidc: failed to decode provider discovery object: %v", err)
 	}
+
 	if p.Issuer != issuer {
 		return nil, fmt.Errorf("oidc: issuer did not match the issuer returned by provider, expected %q got %q", issuer, p.Issuer)
 	}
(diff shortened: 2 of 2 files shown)


oidc.go
```diff
@@ -101,8 +102,14 @@ func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
 		return nil, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 	defer resp.Body.Close()
+
 	var p providerJSON
 	if err := json.Unmarshal(body, &p); err != nil {
+		var isJSON bool
+		content, _, _ := mime.ParseMediaType(r.Header.Get("Content-Type"))
+		if content != "application/json" {
+			err = fmt.Errorf("response Content-Type was not JSON: %v", err)
```
ericchiang: We should include the actual content type in the error message.

oidc.go
```diff
@@ -101,8 +102,14 @@ func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
 		return nil, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 	defer resp.Body.Close()
+
 	var p providerJSON
 	if err := json.Unmarshal(body, &p); err != nil {
+		var isJSON bool
+		content, _, _ := mime.ParseMediaType(r.Header.Get("Content-Type"))
```
ericchiang: Should we be smart about the err value here? e.g.

if err == nil && content != "application/json"? {

oidc.go
chancez: Good idea.

oidc.go
chancez: I don't think so, we only get an error if it cannot parse the media type, which means `content` won't be anything we're explicitly looking for, it will just be "". Since we're providing our own error message, I don't think we need the media type parsing error.

chancez: Although, we might need to check the error case based on your comment below, in order to correctly provide the content-type in the error message. I'll update.

ericchiang: I meant, we should only replace the error if we can detect the content type. But either way, more verbose errors are good.

chancez:
@ericchiang Updated. I'm not 100% happy that the reading of the body has to be done outside the function, or otherwise the unmarshalResp function has to return the body, because there are error cases that want the body to be returned in the error message.

oidc.go
```diff
@@ -307,3 +313,16 @@ func (j *jsonTime) UnmarshalJSON(b []byte) error {
 	*j = jsonTime(time.Unix(unix, 0))
 	return nil
 }
+
+func unmarshalResp(r *http.Response, body []byte, v interface{}) error {
+	err := json.Unmarshal(body, &v)
+	if err != nil {
+		ct := r.Header.Get("Content-Type")
+		mediaType, _, _ := mime.ParseMediaType(ct)
```
ericchiang: you're not capturing the error here.

oidc.go
```diff
@@ -307,3 +313,16 @@ func (j *jsonTime) UnmarshalJSON(b []byte) error {
 	*j = jsonTime(time.Unix(unix, 0))
 	return nil
 }
+
+func unmarshalResp(r *http.Response, body []byte, v interface{}) error {
+	err := json.Unmarshal(body, &v)
+	if err != nil {
```
ericchiang: nit:

```
err := json.Unmarshal(body, &v)
if err == nil {
    // return nil
}
// look at error
```

oidc.go
chancez: I took a look, and the error isn't useful for this. You basically just get errors like `mime: invalid media parameter`, and mediaType is `""` on errors. The error isn't useful to return or add to the error message, and mediaType will still != "application/json" in the error case, thus returning the other error message about failing to parse the media type.

oidc.go
chancez: Fixed.

oidc.go
ericchiang: I'm saying, I'd expect the if condition to look like

```go
	if err == nil && mediaType != "application/json" {
		return fmt.Errorf("expected Content-Type = application/json, got %q", ct)
	}
```

chancez: @ericchiang The extra condition is unnecessary though, it's the same logic with, or without it.

We don't need to check the error because if the `err != nil`, then `mediaType != "application/json"` (because `mediaType == ""` in error cases, so the outcome is the same. I don't feel too strongly about it, I can add it if you like, but it just seems like unnecessary extra noise.

oidc.go
```diff
@@ -307,3 +313,16 @@ func (j *jsonTime) UnmarshalJSON(b []byte) error {
...
+	err := json.Unmarshal(body, &v)
+	if err == nil {
+		return nil
+	}
+	ct := r.Header.Get("Content-Type")
+	mediaType, _, _ := mime.ParseMediaType(ct)
+	if mediaType != "application/json" {
+		return fmt.Errorf("expected Content-Type = application/json, got %q", ct)
+	}
+	return fmt.Errorf("expected Content-Type = application/json, but could not parse media type out of %q: %v", ct, err)
```
ericchiang: This is the wrong error. If you failed to parse the media type, `mediaType` would be "". This can only be hit when you actually correctly parse the media type and it's "application/json"

chancez: Ah yes. I pushed up a new commit fixing this up and changing some of the conditionals to make a little more sense.

oidc.go
```diff
@@ -93,18 +94,23 @@ func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to read response body: %v", err)
```
pquerna: - can this include the source url?
- most other errors from this package include an `oidc:` prefix, not sure if thats expected going forward, but should probably have it for consistency?

jwks.go
```diff
@@ -182,14 +181,16 @@ func (r *remoteKeySet) updateKeys() ([]jose.JSONWebKey, time.Time, error) {
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return nil, time.Time{}, fmt.Errorf("oidc: read response body: %v", err)
+		return nil, time.Time{}, fmt.Errorf("unable to read response body: %v", err)
```
pquerna: - can this include the source url?
- most other errors from this package include an `oidc:` prefix, not sure if thats expected going forward, but should probably have it for consistency?

oidc.go
chancez: @pquerna the caller of this function will wrap this error, and will be the one adding the `oidc:` prefix. I'm not usre it's this functions' job to handle putting the source URL into the error, I believe caller's should do that if they're going to.

PR #149, opened 2017-05-24
oidc: Expose Issuer on the OIDC Provider struct

pquerna:
- I have a couple places where i'm keeping a Provider around in other structs, and I'd like to get at the original Issuer, seems easy to just expose this as another method?

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -143,6 +143,11 @@ func (p *Provider) Endpoint() oauth2.Endpoint {
 	return oauth2.Endpoint{AuthURL: p.authURL, TokenURL: p.tokenURL}
 }
 
+// Issuer returns the OAuth2 issuer claim for the given provider.
+func (p *Provider) Issuer() string {
+	return p.issuer
+}
+
 // UserInfo represents the OpenID Connect userinfo claims.
 type UserInfo struct {
 	Subject       string `json:"sub"`


ericchiang:
@pquerna we exported it previously. Want to just do that?

pquerna:
I'm easy, it seemed intentional that *Provider didn't export any fields, so was jus trying to keep the pattern

ericchiang:
> it seemed intentional that *Provider didn't export any fields,

Yep, I'd just prefer fields to accessers :) I think we're fine to unexport any required metadata claims (just like the IDToken).

oidc.go
```diff
@@ -143,6 +143,11 @@ func (p *Provider) Endpoint() oauth2.Endpoint {
 	return oauth2.Endpoint{AuthURL: p.authURL, TokenURL: p.tokenURL}
 }
 
+// Issuer returns the OAuth2 issuer claim for the given provider.
+func (p *Provider) Issuer() string {
+	return p.issuer
```
ericchiang: As a follow up, can we just make this field exported?

ericchiang:
Going to close this for now. Workaround seems like:

```go
type provider struct {
    *oidc.Provider

    IssuerURL string
}
```

Issue #136, opened 2017-03-06
Failing to parse time when `Cache-Control: max-age` and `Expires` headers are `0`


frodenas:
I'm trying to setup my kubernetes cluster to use [oidc](https://kubernetes.io/docs/admin/authentication/#configuring-the-api-server), but I'm running into some problems.

When the apiserver tries to fetch the discovery data I get this error:

```
oidc.go:178] oidc authenticator: failed to fetch provider discovery data: parsing time "0" as "Mon, 02 Jan 2006 15:04:05 MST": cannot parse "0" as "Mon"
```
Looking at the code, it seems it's failing at `go-oidc` library at the [http.expires](https://github.com/coreos/go-oidc/blob/fe7346e2e6855bda769d5b3db7d917f159c5613c/http/http.go#L89) method when it tries to parse the `Expires` header (as the `Cache-Control: max-age` header is [0](https://github.com/coreos/go-oidc/blob/fe7346e2e6855bda769d5b3db7d917f159c5613c/http/http.go#L79)).

These are my `.well-known/openid-configuration` response headers:

```
< HTTP/1.1 200 OK
* Server Apache-Coyote/1.1 is not blacklisted
< Server: Apache-Coyote/1.1
< Cache-Control: no-cache, no-store, max-age=0, must-revalidate
< Pragma: no-cache
< Expires: 0
< Strict-Transport-Security: max-age=31536000 ; includeSubDomains
< X-XSS-Protection: 1; mode=block
< X-Frame-Options: DENY
< X-Content-Type-Options: nosniff
< Content-Type: application/json;charset=UTF-8
< Transfer-Encoding: chunked
< Date: Mon, 06 Mar 2017 20:39:10 GMT
```

First I thought it was a misconfiguration problem within my server, but looking at the [HTTP/1.1 RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.21), it states: 

*HTTP/1.1 clients and caches MUST treat other invalid date formats, especially including the value "0", as in the past (i.e., "already expired")*

Can you please help me confirm if this is the issue? If yes, would it be possible to accept `0` as a valid `Expires` and return the discovery data as no cacheable? Thanks.

ericchiang:
[...]

PR #153, opened 2017-07-11
fix(http): Allows 0 as an `Expires` header value

thomastaylor312:
This is allowed by the RFC and is common with a few OIDC providers.

Partially addresses #136 as a temporary solution until k8s uses the top level package

/cc @ericchiang


thomastaylor312:
As a note I also manually validated that this is working in Kubernetes 1.7

ericchiang:
lgtm, thanks!

Issue #156, opened 2017-09-05
Fail to call NewProvider()


huikang:
Hi, I am new to OIDC. When I use the go-oidc to connect my provider, it throws the following error

```
oidc: failed to decode provider discovery object: expected Content-Type = application/json, got "text/html": invalid character '<' looking for beginning of value
```
Does the library support other content-type? If so could someone guide how to do so? Thanks.

ericchiang:
It looks like your provider is returning HTML to it's JSON discovery endpoint. This is was a correct response looks like https://accounts.google.com/.well-known/openid-configuration

Are you sure your issuer URL is correct? For example, for google I'd have to enter "https://accounts.google.com" and then the rest of the URL is implied.

huikang:
Hi, @ericchiang I am connecting to an internal provider. But I am not sure if _text/html_ is the only option. Does go-oidc support other content type?

ericchiang:
> Does go-oidc support other content type?

The spec requires JSON. https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse

The content type actually isn't enforced by go-oidc. It's just mentioned if the library fails to parse the response from the provider.

huikang:
@ericchiang you are right; it should be json format. I consulted with our internal OIDC provider team and learned that they do not support OIDC Discovery as of now. Therefore,  I need to manually configure the endpoints. Could you point me to some example using your go-oidc library? Thanks.

ericchiang:
@huikang this package currently doesn't work with providers that don't support OIDC discovery. There's been some previous discussion previously though https://github.com/coreos/go-oidc/issues/155#issuecomment-322856151

Maybe one of those proposed API changes work for you?

Issue #155, opened 2017-08-14
Issuer URL is not normalized for comparison


voutasaurus:
The NewProvider function takes an issuer argument and creates the config URL from it with
```
wellKnown := strings.TrimSuffix(issuer, "/") + "/.well-known/openid-configuration"
```

Then later the issuer is checked against that returned by the config endpoint.
```
if p.Issuer != issuer {
```

If an issuer of `https://issuer.com/` is passed in but the config endpoint returns `https://issuer.com` as the issuer, then this comparison will fail even though these are clearly the same issuer.

I suggest normalizing the issuer url for the comparison (e.g. like this):
```
if strings.TrimSuffix(p.Issuer, "/") != strings.TrimSuffix(issuer, "/") {
```

Does this sound appropriate? If so I can make the change and PR it.

ericchiang:
> Does this sound appropriate?

Issuers are exact strings. `https://issuer.com/` is unique from `https://issuer.com` and those two aren't equal. We do the `TrimSuffix` is because the spec mandates i:

> If the Issuer value contains a path component, any terminating / MUST be removed before appending /.well-known/openid-configuration.

https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest

However, the issuer value returned by the metadata must exactly match the original issuer string.

This check is strict, but often helps us identify mis-configurations (https://github.com/coreos/go-oidc/issues/154, https://github.com/coreos/go-oidc/issues/121). Probably better for us to make the docstrings stress this more. But I don't think we should do any normalization during comparison.

voutasaurus:
Both of those mis-configurations would be still be caught if the terminating `/` was removed for the comparison.

The normalization is especially annoying because some issuers have terminating `/`s and some don't. So I can't just strip them or append them myself.

As a result without the normalization in the check itself I will be calling the `/.well-known/openid-configuration` endpoint twice. Once to get the exact issuer string and then again in NewProvider. If I don't do this myself then my users will get errors if they normalize their URLs in a way that doesn't exactly match what the json config for their provider is returning.

ericchiang:
It seems odd to do this normalization here, but not elsewhere. I could provide an issuer URL `https://issuer.com/` then later validated an ID token when it has an issuer URL `https://issuer.com`. That breaks a lot of assumptions about the issuer URL being unique across a program.

It also seems odd to be using this package just to validate issuer URLs.

[...]

PR #161, opened 2017-10-20
*: expose KeySet, NewRemoteKeySet, and NewVerifier

ericchiang:
Expose internal types to let users create IDTokenVerifiers without
using metadata discovery (/.well-known/openid-configuration). This
expands support to providers that don't implement discovery, and
lets users deliver verification keys out-of-band.

Closes https://github.com/coreos/go-oidc/issues/156
Addresses concerns brought up in https://github.com/coreos/go-oidc/issues/155

assigning @rithujohn191 
cc @pquerna in case you're interested in using the newly exposed types.


jwks.go
```diff
@@ -23,6 +23,20 @@ import (
 // updated.
 const keysExpiryDelta = 30 * time.Second
 
+// NewRemoteKeySet returns a KeySet that can validate JSON web tokens by using HTTP
+// GETs to fetch JSON web token sets hosted at a remote URL. This is automatically
+// use by NewProvider using the URLs returned by OpenID Connect discovery, but is
```
rithujohn191: use -> used

verify.go
```diff
@@ -19,19 +19,41 @@ const (
...
-// testing. Outside of testing, it's always backed by a remoteKeySet.
-type keySet interface {
-	verify(ctx context.Context, jws *jose.JSONWebSignature) ([]byte, error)
+// KeySet is a set of publc JSON web keys that can be used to validate the signature
+// of JSON web tokens. This is expected to be backed by a remote key set through
+// provider metadata discovery or an in-memory set of keys delivered out-of-band.
+type KeySet interface {
+	// VerifySignature parses the JSON web token, verifies the signature, and returns
+	// the raw payload. Header and claim fields are validated by other parts of the
+	// package. The KeySet does not need to check values such as signature algorithm,
```
rithujohn191: is there a reason you mentioned this specifically? are the rest of the values checked somewhere else?

rithujohn191 (commented):
It would really help if you could have a an example that demos the use of it

verify.go
ericchiang: The IDTokenVerifier already performs these checks. Going to update the docstring:

```
        // VerifySignature parses the JSON web token, verifies the signature, and returns                              
        // the raw payload. Header and claim fields are validated by other parts of the
        // package. For example, the KeySet does not need to check values such as signature 
        // algorithm, issuer, and audience since the IDTokenVerifier validates these values
        // independently.                                  
```

Does that work?

verify.go
```diff
@@ -19,19 +19,54 @@ const (
...
 	issuer string
 }
 
+// NewVerifier returns a verifier manually constructed from a key set and issuer URL.
+//
+// It's easier to use provider discovery to construct an IDTokenVerifier than creating
+// one directly. This method is intended to be used with provider that don't support
+// metadata discovery, or avoiding round trips when the key set URL is already known.
+//
+// This constructor can be used to create a verifier directly using the issuer URL and
```
ericchiang, line 50: @rithujohn191 does this look better?

rithujohn191 (approved):
LGTM

PR #165, opened 2018-01-16
NewVerifier: fix defaulting of RS256

ericchiang:
When NewVerifier was introduced, it forgot to default the
SupportedSigningAlgs value in the verification config. This means
an attacker can pass a token signed with any asymmetric "alg" value.

This isn't a P0 because the public key set from the provider should
only return asymmetric keys, so an attacker can't sneak a token
signed with a symmetric algorithm like HS512. RS256 is also the
weakest hash supported by square/go-jose, so you can't downgrade to
a weaker signing hash. Additionally, using jose.ParseSigned ensures
tokens encrypted with algorithms like A128GCM are rejected.

Additionally NewVerifier isn't expected to be used commonly. It's
mostly for testing (though that doesn't reduce the severity).

Unify the verifier creation code and make it impossible to pass an
empty list of SupportedSigningAlgs.

No new tests because the Verify path is already tested.

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -94,21 +94,7 @@ type Config struct {
 // The returned IDTokenVerifier is tied to the Provider's context and its behavior is
 // undefined once the Provider's context is canceled.
 func (p *Provider) Verifier(config *Config) *IDTokenVerifier {
-
-	return newVerifier(p.remoteKeySet, config, p.issuer)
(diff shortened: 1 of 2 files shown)


rithujohn191 (approved):
LGTM

PR #168, opened 2018-04-19
verify: added skip verify issuer

slot:
In order to be able to work with misconfigured auth servers I added this config flag to disable the issuer check. By default issuer check will be enabled. The special handling for google servers remains untouched.


slot:
wrong target branch

PR #169, opened 2018-04-19
verify: added skip verify issuer

slot:
In order to be able to work with misconfigured auth servers I added this config flag to disable the issuer check. By default issuer check will be enabled. The special handling for google servers remains untouched.

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -49,6 +49,8 @@ type Config struct {
 	SkipClientIDCheck bool
 	// If true, token expiry is not checked.
 	SkipExpiryCheck bool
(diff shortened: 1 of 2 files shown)


ericchiang:
I'd really not like to add this. Issuer verification is fundamental to this validation. If you don't want to do it you can already parse the JWT yourself.

Which servers are misconfigured? Can you just configure them correctly instead?

ericchiang:
Closing. This isn't something we want to be configurable since it indicates a mis-configuration that should really be fixed.

slot:
Agreed. Thanks for considering.

Issue #171, opened 2018-05-24
Distributed claims utilities


ericchiang:
http://openid.net/specs/openid-connect-core-1_0.html#AggregatedDistributedClaims

ref https://github.com/kubernetes/kubernetes/pull/63213 

Consider adding some sort of distributed claim resolver. Maybe another field on the IDToken?

```
type IDToken struct { }

// Stays the same, doesn't do any resolution
func (t *IDToken) Claims(into interface{}) error {}

// If the named claim is a distributed claim, this will be resolved and unmarshalled into "v"
func (t *IDToken) Claim(name string, v interface{}) error {}
```

Claim expansion could be done when calling `IDToken.Claim` or could be built into `IDTokenVerifier.Verify`

cc @filmil for any thoughts here

ericchiang:
cc @sean-q-sun

filmil:
Hi folks.  I was away from keyboard for a few days, hence the late response.

This certainly seems like an useful addition to have out of the box.

A few questions come to mind:

- What is the roll-out strategy for the distributed claims?  Should they become available to the users out of the box?  (I mean, what is the incentive to start using it if it requires a code change?)
- Would we also want to support aggregated claims on the library level?  If yes, is there a benefit to keeping the structure of the claim somehow visible (e.g. if there are conflicting claims, do we expose all of them and how)
- I think that the OIDC spec is somewhat unclear about the structure of the distributed claim response.  E.g. it seems that while the response to the claim must be a JWT, it says nothing about whether it may be wrapped into something.

Hope this helps, even if belated, as you've already started working on it.

Cheers,
F

ericchiang:
> Hi folks. I was away from keyboard for a few days, hence the late response.

Hope you enjoyed your time off!

> What is the roll-out strategy for the distributed claims? Should they become available to the users out of the box? (I mean, what is the incentive to start using it if it requires a code change?)

The proposed change is to keep all the existing behavior the same, but introduce a new method `Claim` to the `IDToken` struct:

```go
// Claim unmarshals the value of the named claim into v.
//
//    var email string
//    if err := idToken.Claim("email", &email); err != nil {
//        // handle error
//    }
//
// If the named claims is a distributed claim, Claim queries the remote endpoint to
// resolve the claim value.
//
// See: https://openid.net/specs/openid-connect-core-1_0.html#DistributedExample
func (i *IDToken) Claim(name string, v interface{}) error
```

We'd also add something to the verifier config:

```go
[...]

PR #172, opened 2018-05-25
Resolve distributed claims in idToken

seanqsun:
Updates #171
Needs more test cases and functionality. Right now, it only parses out the claims from the ID Token and puts them in `distributedClaims`

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -155,15 +155,30 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		return nil, fmt.Errorf("oidc: failed to unmarshal claims: %v", err)
 	}
 
+	distributedClaims := make(map[string]claimSource)
+
+	//step through the token to map claim names to claim sources"
+	for cn, src := range token.ClaimNames {
+		if src == "" {
+			return nil, fmt.Errorf("oidc: failed to obtain source from claim name")
+		}
+		s, ok := token.ClaimSources[src]
+		if !ok {
+			return nil, fmt.Errorf("oidc: source does not exist")
+		}
+		distributedClaims[cn] = s
+	}
+
 	t := &IDToken{
-		Issuer:          token.Issuer,
-		Subject:         token.Subject,
-		Audience:        []string(token.Audience),
-		Expiry:          time.Time(token.Expiry),
-		IssuedAt:        time.Time(token.IssuedAt),
-		Nonce:           token.Nonce,
-		AccessTokenHash: token.AtHash,
-		claims:          payload,
+		Issuer:            token.Issuer,
+		Subject:           token.Subject,
+		Audience:          []string(token.Audience),
+		Expiry:            time.Time(token.Expiry),
+		IssuedAt:          time.Time(token.IssuedAt),
+		Nonce:             token.Nonce,
+		AccessTokenHash:   token.AtHash,
+		claims:            payload,
+		distributedClaims: distributedClaims,
 	}
 
 	// Check issuer.
--- a/verify_test.go
+++ b/verify_test.go
@@ -3,6 +3,7 @@ package oidc
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


verify_test.go
```diff
@@ -216,6 +217,147 @@ func TestAccessTokenHash(t *testing.T) {
...
+				name: "Distributed claims match. 1 name : 1 source",
+				idToken: `{
+							"iss":"https://foo","aud":"client1",
+							  "_claim_names": {
+							    "address": "src1"
+						 	  },
+						 	  "_claim_sources": {
+							    "src1": {"endpoint": "123", "access_token":"1234"}
+							  }
+						  }`,
```
ericchiang: Formatting here looks wonky. Maybe try to clean this up a little?

verify_test.go
```diff
@@ -216,6 +217,147 @@ func TestAccessTokenHash(t *testing.T) {
...
 }
 
+func TestDistributedClaims(t *testing.T) {
+	tests := []struct {
+		test verificationTest
+		want map[string]claimSource
+	}{
+		{
+			test: verificationTest{
+				name:    "No distributed claims",
```
ericchiang: Subtest names are usually shorter and don't contain spaces. Maybe "NoDistClaims"?

verify_test.go
```diff
@@ -216,6 +217,147 @@ func TestAccessTokenHash(t *testing.T) {
...
+					ClientID:        "client1",
+					SkipExpiryCheck: true,
+				},
+				signKey: newRSAKey(t),
+			},
+			want: map[string]claimSource{},
+		},
+		{
+			test: verificationTest{
+				name: "Distributed claims match. 1 name : 1 source",
```
ericchiang: "1DistClaim"

verify_test.go
```diff
@@ -216,6 +217,147 @@ func TestAccessTokenHash(t *testing.T) {
...
+			test: verificationTest{
+				name: "Distributed claims match. 1 name : 2 sources",
+				idToken: `{
+							"iss":"https://foo","aud":"client1",
+							  "_claim_names": {
+							    "address": "src1"
+						 	  },
+						 	  "_claim_sources": {
+								"src1": {"endpoint": "https://foo", "access_token":"1234"},
+								"src1": {"endpoint": "https://bar", "access_token":"1234"}
```
ericchiang: This seems like an ambiguous test since the endpoint could be "https://foo" or "https://bar" probably better to drop this test case.

verify_test.go
```diff
@@ -216,6 +217,147 @@ func TestAccessTokenHash(t *testing.T) {
...
+						 	  "_claim_sources": {
+							  }
+						  }`,
+				config: Config{
+					ClientID:        "client1",
+					SkipExpiryCheck: true,
+				},
+				signKey: newRSAKey(t),
+			},
+			want: map[string]claimSource{
```
ericchiang: This should result in an error, right?

verify_test.go
```diff
@@ -216,6 +217,147 @@ func TestAccessTokenHash(t *testing.T) {
 	})
 }
 
+func TestDistributedClaims(t *testing.T) {
+	tests := []struct {
+		test verificationTest
+		want map[string]claimSource
```
ericchiang: Add a `wantErr` bool here to indicate that we want the ID token to fail parsing.

verify.go
```diff
@@ -155,15 +155,25 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		return nil, fmt.Errorf("oidc: failed to unmarshal claims: %v", err)
 	}
 
+	distributedClaims := make(map[string]claimSource)
+
+	//step through the token to mp claim names to claim sources"
+	for cn, src := range token.ClaimNames {
+		if src != "" {
```
ericchiang: probably

```
if src == "" {
    // return error
}
s, ok := token.ClaimsSources[src]
if !ok {
    // return error because source name doesn't exist
}
distributedClaims[cn] = s
```

ericchiang (commented):
I think we should change the runGetToken signature to return an error. That way instead of checking if this idToken is nil we can check the returned error.

```
func (v verificationTest) runGetToken(t *testing.T) (*IDToken, error) {
    // ...
}

func (v verificationTest) run(t *testing.T) {
    err := v.runGetToken(t)
    if err != nil && !v.wantErr {
        t.Errorf("%v", err)
    }
    if err == nil && v.wantErr {
        t.Errorf("expected error")
    }
}
```

ericchiang:
Also having `Fixes #171` in your PR comment will close the issue. You probably want `Updates #171` or something like that :)

ericchiang (approved):
lgtm. squash and we'll merge it

PR #173, opened 2018-05-31
Verify distributed claim endpoints

seanqsun:
Updates #171

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -7,6 +7,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


verify_test.go
```diff
@@ -358,10 +452,45 @@ type verificationTest struct {
 	// testing invalid signatures.
 	verificationKey *signingKey
 
-	config  Config
+	config Config
+
```
ericchiang: please remove this whitespace

verify_test.go
```diff
@@ -342,6 +345,97 @@ func TestDistributedClaims(t *testing.T) {
...
+func TestDistClaimResolver(t *testing.T) {
+	tests := []struct {
+		test    resolverTest
+		want    map[string]claimSource
+		wantErr bool
+	}{
+		{
+			test: resolverTest{
+				name:    "noAccessToken",
+				payload: `{"iss":"","aud":"client1","email":"janedoe@email.com"}`,
```
ericchiang: include an issuer in these tests

verify_test.go
```diff
@@ -342,6 +345,97 @@ func TestDistributedClaims(t *testing.T) {
 	}
 }
 
+func TestDistClaimResolver(t *testing.T) {
+	tests := []struct {
```
ericchiang: This should just be

```
tests := []resolverTest{
  // ...
}
```

We had the `test.test` in the other ones as a hack :)

verify_test.go
```diff
@@ -342,6 +345,97 @@ func TestDistributedClaims(t *testing.T) {
...
+				accessToken: "12345",
+			},
+			wantErr: true,
+		},
+	}
+	for _, test := range tests {
+
+		t.Run(test.test.name, func(t *testing.T) {
+
+			endpoint, err := test.test.testEndpoint(t)
```
ericchiang: this doesn't return an endpoint? does it? Need to rename it as "claims"

Also please remove the whitespace in the two lines above.

verify.go
```diff
@@ -118,6 +120,54 @@ func contains(sli []string, ele string) bool {
 	return false
 }
 
+// Returns the Claims from the distributed JWT token
+func resolveDistributedClaim(ctx context.Context, verifier *IDTokenVerifier, src claimSource) ([]byte, error) {
+
```
ericchiang: no whitespace to start an indented block.

verify.go
```diff
@@ -118,6 +120,54 @@ func contains(sli []string, ele string) bool {
...
+	req, err := http.NewRequest("GET", src.Endpoint, nil)
+	if err != nil {
+		return nil, fmt.Errorf("oidc: malformed request: %v", err)
```
ericchiang: nit: this function isn't exported, so something else in the package going to prefix these errors with `oidc:`, we can remove this here.

verify.go
```diff
@@ -118,6 +120,54 @@ func contains(sli []string, ele string) bool {
...
+
+	token, err := verifier.Verify(ctx, string(body))
+	if err != nil {
+		return nil, fmt.Errorf("malformed response body: %v", err)
+	}
+
+	return token.claims, nil
+}
+
+func parseClaim(raw []byte, name string, v interface{}) error {
```
ericchiang, line 156: add a test for this method?

ericchiang (approved):
lgtm

PR #176, opened 2018-08-27
jwks: Don't store context in remoteKeySet

bitglue:
The docs for the context package say:

    Do not store Contexts inside a struct type; instead, pass a Context
    explicitly to each function that needs it.

Previously the context passed to NewProvider would get stashed away in
the remoteKeySet structure, then maybe used later. If the context that
created the provider had been cancelled, keys could not be fetched.

For a use case where this is bad, consider the user of this library is
an HTTP server. It has a REST endpoint which can configure OIDC
providers to use for authentication. It then has other endpoints that
use those providers. When the configure endpoint is called, it creates a
context and calls NewProvider(). When the handler for the configure
endpoint is done, it cancels the context.

When the provider returned by NewProvider() is subsequently used, keys
can not be verified: they fail with "context canceled".

Fortunately the keys get updated from Verify(), which already takes a
context. So we can just use that. This does mean if the context is being
used to change the default HTTP client it's necessary to do it on the
context passed to Verify() as well as NewProvider().

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -31,22 +31,19 @@ const keysExpiryDelta = 30 * time.Second
 //
 // The returned KeySet is a long lived verifier that caches keys based on cache-control
 // headers. Reuse a common remote key set instead of creating new ones as needed.
-//
-// The behavior of the returned KeySet is undefined once the context is canceled.
-func NewRemoteKeySet(ctx context.Context, jwksURL string) KeySet {
-	return newRemoteKeySet(ctx, jwksURL, time.Now)
+func NewRemoteKeySet(jwksURL string) KeySet {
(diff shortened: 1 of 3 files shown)


ericchiang (commented):
While I'm sympathetic to this change, it's as a breaking change to the API so we can't make it before a v3 branch.

The context for the NewProvider is expected to indicate the lifetime of the remote key set. It's so we can eventually shut down the goroutine that's created globally to the set. A Close method might work better, it's the way the API works today.

bitglue:
Maybe you could consider this a documentation bug then?

ericchiang:
Yep that works for me

PR #177, opened 2018-08-31
DecodedClaimsJSON endpoint for userinfo

ajmyyra:
New DecodedClaimsJSON() endpoint for UserInfo. Endpoint returns the JSON string as base64-encoded json.RawMessage, to easily add userinfo as header for other service endpoints.

As most [standard claims](http://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims) aren't included in UserInfo struct, and the OIDC standard allows people also to define their own, UserInfo.claims is the most useful part of the userinfo model, so it's good to offer a simple endpoint for fetching the data.

If needed, I can make suggested changes to the code (if you wish to have a different name for it, for example "Base64ClaimsJSON()"), or you can freely do so yourselves if you want to.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -176,6 +176,19 @@ func (u *UserInfo) Claims(v interface{}) error {
 	return json.Unmarshal(u.claims, v)
 }
 
+// EncodedClaimsJSON returns claims JSON in base64-encoded form
+func (u *UserInfo) EncodedClaimsJSON() (json.RawMessage, error) {
+	if u.claims == nil {
+		return nil, errors.New("oidc: claims not set")
+	}
+	cj, err := json.Marshal(u.claims)
+	if err != nil {
+		return nil, errors.New("oidc: problem marshalling claims to JSON: " + err.Error())
+	}
+
+	return cj, nil
+}
+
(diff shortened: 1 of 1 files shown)


ericchiang:
Thanks for the PR!

It seems like you can already do everything you're describing without adding new API. To get the raw claims:

```
userInfo, err := provider.UserInfo(ctx, token)
if err != nil {
    // handle error
}
raw := json.RawMessage{}
if err := userInfo.Claims(&raw); err != nil {
    // handle error
}
```

Or if you've defined your own user info:

```
type MyUserInfo struct {
    // fields
}

// ...

userInfo, err := provider.UserInfo(ctx, token)
if err != nil {
    // handle error
}
myUserInfo := myUserInfo{}
if err := userInfo.Claims(&myUserInfo); err != nil {
    // handle error
}
```

It might not be intuitive but it seems better than adding two ways to do the same thing. Maybe update the docs instead of adding another method?

ajmyyra:
Seems I missed that completely. Like you said, no need to add another way to do it. Sorry for the wasted time and thank you for guidance!

PR #178, opened 2018-09-03
Remove client_secret POST parameter for basic auth method

jangaraj:
My custom IdP returns error, when I want to exchange a code for the access token:
```
invalid_request: Request requires a valid client authentication method
```
Root cause: `go-oidc v1` sends 2 auth methods (post+basic auth) in one request - it is a violation of https://tools.ietf.org/html/rfc6749#section-2.3

> the client MUST NOT use more than one authentication method in each request

More details about the issue: https://github.com/gambol99/keycloak-proxy/issues/336

### DIFF
--- a/oauth2/oauth2.go
+++ b/oauth2/oauth2.go
@@ -257,7 +257,6 @@ func (c *Client) RequestToken(grantType, value string) (result TokenResponse, er
 	v := c.commonURLValues()
 
 	v.Set("grant_type", grantType)
-	v.Set("client_secret", c.creds.Secret)
 	switch grantType {
 	case GrantTypeAuthCode:
 		v.Set("code", value)


ericchiang:
The v1 branch has been marked as depreciated for over a year and we've
stopped accepting PRs for it. This might be a good time to switch your
project over to the golang.org/x/oauth2 client.

jangaraj:
I agree, but the implementation of `v2` is not a quick task. I have requested `v2` and it is in the long-term TODO list - https://github.com/gambol99/keycloak-proxy/pull/407#issuecomment-409207544
Will be it be possible to fix `v1` for now?

ericchiang:
I want to be consistent with past requests to fix the v1 branch. We don't
want to do the work to maintain it and users should switch to something
else. I feel that we've done a fair job of trying to communicate that.

If you'd like to patch v1, please fork it locally.

PR #180, opened 2018-09-29
Add basic support for client registration

darh:


ericchiang:
What providers support this? I've never actually come across one that actually ever does.

client.go
```diff
@@ -0,0 +1,113 @@
...
+	if len(p.registrationURL) == 0 {
+		return nil, fmt.Errorf("Can not perform client registration without 'registration_endpoint'")
+	}
+
+	enc, err := json.Marshal(request)
+	if err != nil {
+		return nil, fmt.Errorf("unable to encode request: %v", err)
+	}
+
+	req, err := http.NewRequest("POST", p.registrationURL, bytes.NewBuffer(enc))
```
ericchiang, line 63: I don't like this request being buried in the package, since it will likely require authentication credentials. This seems like this functionality would be better as an example, instead of part of go-oidc directly.

darh: I understand. This (request code/functionality) is not much different than the `NewProvider()`in `oidc.go`. Having this code here, as a part of the package, it allows us to extend `providerJSON` and `Provider` and have the `RegistrationURL` and others filled by unmarshal func.

darh:
Until this neither have I. It is used by identity provider we're integrating with (https://daasi.de/) and since it is part of the OIDC standard I wanted to extend this package. Service is not public yet.

client.go
ericchiang: The endpoints hit by `NewProvider` are assumed to be public and require no additional credentials.

darh: Is this a question? :)

Yes. This solves a problem when no cedentials are required. `ClientRegistration` struct can/could be extended to support that as well.

PR #184, opened 2018-10-31
Add HTTPClient interface to support different HTTP clients

Bezbran:


ericchiang:
This is a breaking change to the public API, so we can't do this here. Also you can already switch out the client transport, which does the exact same thing as this PR.

https://golang.org/pkg/net/http/#Client
https://golang.org/pkg/net/http/#RoundTripper

Bezbran:
Thanks for your response.
You are right about the breaking changes and the ability to switch the
transport of the http.client.
I've found that in some none-basic cases it usefull to replace the Do
function of the http.client.
I have an oauth2 server behind reverse proxy, so I need to send the "Host"
header in each HTTP request to this server.
I made this change to the API so I can add custom headers to the requests.
This problem is classic for interfaces implementation and have no solution
by replacing the client transport.

‫בתאריך יום ד׳, 31 באוק׳ 2018 ב-18:51 מאת ‪Eric Chiang‬‏ <‪
notifications@github.com‬‏>:‬

ericchiang:
This code does what you're describing without changes to the package :)

```go
type myRoundTripper struct {
    rt   http.RoundTripper
    host string
}

func (rt *myRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
    req.Header.Set("Proxy-To-Host", rt.host)
    return rt.rt.RoundTrip(req)
}

func contextForHost(ctx context.Context, host string) context.Context {
    return oidc.ClientContext(ctx, &http.Client{
        Transport: &myRoundTripper{
            rt:   http.DefaultTransport,
            host: host,
        },
    })
}
```

Bezbran:
TNX a lot.

Issue #181, opened 2018-10-22
The code unnecessarily set client_secret 


SongGithub:
## direct consequence of the problem:
I was trying to integrate [Keycloak](https://github.com/keycloak/keycloak-gatekeeper) with Microsofts' implementation of OIDC, and getting 403 error that is ```"keycloak-gatekeeper/handlers.go:121","msg":"unable to exchange code for access token","error":"invalid_request: MSIS9631: Received invalid OAuth request. Multiple client authentication methods were attempted.```

That was because MS's OIDC protocol mandates [The client MUST NOT use more than one authentication method in each request](https://tools.ietf.org/html/rfc6749#section-2.3), and Keycloak was sending auth-code to exchange for `access token` at ADFS server's token endpoint, with both client-authentication methods: `basic` and `post`, which violates the above mentioned mandate. 

I read through Keycloak's code and trace it down to this repo which referred as a go-library. And I found out that when using Keycloak's default auth method `basic auth`, it should only set security header, not to have client_secret in the HTTP POST body as well. Unfortunately, there is a line of unnecessary code that set client_secret in a variable that end up in HTTP POST body before the code could determine which auth method was the calling RP client set to using. 

## cause in the code
[This line of code](https://github.com/coreos/go-oidc/blob/e860bd55bfa7d7cb35d30d26a167982584f616b0/oauth2/oauth2.go#L260) is totally unnecessary, as the client_secret will be set in [downstream method](https://github.com/coreos/go-oidc/blob/e860bd55bfa7d7cb35d30d26a167982584f616b0/oauth2/oauth2.go#L185) only when it is using method AuthMethodClientSecretPost. 

In current setup, the other auth method AuthMethodClientSecretBasic will have client_secret in the Body which is not tolerated by MS's OIDC (ADFS/OIDC/Token-endpoint)

SongGithub:
I can help to make a PR to fix the problem. please let me know. thanks.

ericchiang:
The code you're referring is the v1 branch, which was mark deprecated a year ago. The v2 branch uses golang.org/x/oauth2. Keycloak needs to switch.

SongGithub:
indeed. thanks for the insight!

abstractj:
@SongGithub this is being tracked here https://issues.jboss.org/browse/KEYCLOAK-7265 an planned for further releases. Go-oidc changed a lot the codebase, so we need to make sure that everything is in place.

PR #189, opened 2018-12-11
fix: handle -1 HTTP Expires header

adrien-f:
Greetings !

While using this library on an Azure AD endpoint, I noticed Microsoft sending me `Expires: -1` in response to the OIDC Discovery endpoint. When I saw that you were handling the `0` case, I thought I'd also add the `-1` case as well.

Let me know what you think.

Thanks :smile:

### DIFF
--- a/http/http.go
+++ b/http/http.go
@@ -93,7 +93,7 @@ func expires(date, expires string) (time.Duration, bool, error) {
 
 	var te time.Time
 	var err error
-	if expires == "0" {
+	if expires == "0" || expires == "-1" {
 		return 0, false, nil
 	}
 	te, err = time.Parse(time.RFC1123, expires)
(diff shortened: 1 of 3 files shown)


ericchiang:
@adrien-f, thanks for opening but the branch you've sent a PR for was deprecated more than a year ago. Similar to other issues, I'd recommend either switching to the v2 branch or maintaining a fork for your project.

https://github.com/coreos/go-oidc/issues/181#issuecomment-431870483

adrien-f:
Hey ! 

No worries I understand, It was mostly as a fix for the upstream project https://github.com/keycloak/keycloak-gatekeeper, not sure if they plan to migrate any time soon. In the meantime I'll keep the fork.

Have a good one !

PR #190, opened 2019-01-22
Added logout functionality through provider

dlouwers:
Hi,

Had a need for `end_session_endpoint ` so added this along with a utility function to add parameters mentioned in the spec.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -69,6 +70,7 @@ type Provider struct {
 	authURL     string
 	tokenURL    string
 	userInfoURL string
+	logoutURL   string
 
 	// Raw claims returned by the server.
 	rawClaims []byte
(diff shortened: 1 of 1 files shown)


bruno-vwdilab:
Hello!
Any news about this? Planning on merging? We're also very interested in this!
Thanks! Good job guys.

oidc.go
```diff
@@ -158,6 +162,27 @@ func (p *Provider) Endpoint() oauth2.Endpoint {
 	return oauth2.Endpoint{AuthURL: p.authURL, TokenURL: p.tokenURL}
 }
 
+// LogoutURL returns the logout endpoints for the given provider.
+// See: https://openid.net/specs/openid-connect-session-1_0.html#RPLogout
+func (p *Provider) LogoutURL(idTokenHint string, redirectURL string, state string) string {
```
ericchiang, line 167: Where do these values come from? E.g. how would a user get the idTokenHint and redirectURL?

ericchiang (commented):
I'm not familiar with this flow. Does the client need to POST to this endpoint with its credentials or something?

Also what providers implement this?

ericchiang:
@bruno-vwdilab fyi you don't need to wait for this to merge to access end_session_endpoint:

```
p, err := oidc.NewProvider(ctx, myIssuer)
if err != nil {
    // handle error
}
var claims struct {
    EndSessionURL string `json:"end_session_endpoint"`
}
    // handle error
}
// use claims.EndSessionURL
```

oidc.go
dlouwers: For `idTokenHint` I am using the token value, which works on Azure AD OpenID Connect. The `redirectURL` I am getting off of the request `url` query parameter, but sources could differ. It must be registered as a valid callback url, just like a login url. This callback makes sure that the application session is ended and can be registered separately as a logout url in Azure so that it calls all registered logout url's when a user has signed off, effectively implementing single signoff.

In my case:
```
http.Redirect(w, r, openIDConfig.provider.LogoutURL(openIDTokenCookie.Value, callbackURL, redirectURL), http.StatusFound)
```
See: https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-protocols-openid-connect-code#single-sign-out

bruno-vwdilab:
> @bruno-vwdilab fyi you don't need to wait for this to merge to access end_session_endpoint:
> 
> ```
> p, err := oidc.NewProvider(ctx, myIssuer)
> if err != nil {
>     // handle error
> }
> var claims struct {
>     EndSessionURL string `json:"end_session_endpoint"`
> }
> if err := p.Claims(&claims); err != nil {
>     // handle error
> }
> // use claims.EndSessionURL
> ```

Perfect @ericchiang 
Thanks for the help.

dlouwers:
@ericchiang Are there plans to merge this or is there something that still needs to be done?

dlouwers:
Closed in favor of #226

PR #194, opened 2019-03-12
fix google issuer claim condition

humboldt-xie:
condition `v.issuer == issuerGoogleAccounts && t.Issuer == issuerGoogleAccountsNoScheme `

is alway false


humboldt-xie:
v.issuer and t.Issuer

Issue #183, opened 2018-10-30
Skip issuer URL check


pankrator:
Would it make sense to add possibility to skip the Issuer URL [check][2], like the [client id][1]?
We have a use case where the Authorization server issues tokens with different issuer urls.

[1]: https://github.com/coreos/go-oidc/blob/v2/verify.go#L86
[2]: https://github.com/coreos/go-oidc/blob/v2/verify.go#L234-L243

ericchiang:
dup of https://github.com/coreos/go-oidc/issues/159 and several other issues

tl;dr - no

Issue #159, opened 2017-10-04
The URL to discover and Issue need not be same


mkhadilk:
In Package oidc and compilation unit oidc.go, "NewProvider (" call implementation is looks wrong.

if **p.Issuer != issuer** {
		return nil, fmt.Errorf("oidc: issuer did not match the issuer returned by provider, expected %q got %q", issuer, p.Issuer)
	}

This code looks wrong to me. I am reading the OIDC discovery code and I dont see why the URL one provides should be the URL of the Issuer.

jmcarp:
I'm running into the same issue. Would you be open to dropping that check or reviewing a patch that drops it @ericchiang ?

ericchiang:
Sorry for the delayed response, have been out of the office.

Copying from a previous thread https://github.com/coreos/go-oidc/issues/154#issuecomment-318493898

> ... the spec section you're looking for is https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig
>
> > OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer.
>
> We've seen this check regularly catch problems when users are referring to an provider through a DNS name alternative to the provider value. e.g. #121

From my reading of the spec these values MUST match.

jmcarp:
Thanks @ericchiang! I think the relevant part of the spec is

> The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve the configuration information. This MUST also be identical to the iss Claim value in ID Tokens issued from this Issuer.

Looks like there's a [bug](https://github.com/cloudfoundry/uaa/pull/672) in the provider we're using--we'll resolve upstream.

chaliy:
In my case, we are building infrastructure for the project and it makes it hard to implement public and private DNS resolutions to be the same. So basically we need a way of having correct issuer and at the same time openid-configuration should be resolvable in private. There are few ways to do this:

1. Tweak private DNS to also resolve public one. This is good approach. In our case it will be hard to implement because of multitenancy. 
2. Make oidc client so that it could have other way to get oidc configuration. For example allow custom URL to fetch /.well-known/openid-configuration

Would you consider to reopen this issue and accept PR with optional way to override wellknow URL?

[...]

PR #195, opened 2019-03-14
add config.SkipIssuerCheck

ti:
config.SkipIssuerCheck is useful for some special projects

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -86,6 +86,8 @@ type Config struct {
 	SkipClientIDCheck bool
 	// If true, token expiry is not checked.
 	SkipExpiryCheck bool
+	// If true, no Issuer check performed.
+	SkipIssuerCheck bool
 
 	// Time function to check Token expiry. Defaults to time.Now
 	Now func() time.Time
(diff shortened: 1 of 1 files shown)


ericchiang:
Hi ti@! This is a pretty common request that I've said no to several times in the past:

https://github.com/coreos/go-oidc/issues/121
https://github.com/coreos/go-oidc/pull/122
https://github.com/coreos/go-oidc/issues/159
https://github.com/coreos/go-oidc/pull/169
https://github.com/coreos/go-oidc/issues/183

If you're seeing an issuer mis-match, it's likely that there's some kind of misconfiguration. The issuer check is also fairly fundamental to the OpenID Connect spec. Which projects are you referring to?

ti:
@ericchiang Thanks for your reply. But in some case, it is not Azure or something else.

for example:
    
A oidc server service, is allowed user to custom domain,  so the issuer is changed. but in these server want to check the idToken, but the issuer is dynamic. if SkipIssuerCheck is on, the auth server can check any idtoken.


i know you have some securiy concern.  but the bussiness always complex。 the SkipExpiryCheck and SkipClientIDCheck are on there anyway.

ti:
@ericchiang Thank you very very much.

ti:
:-D， I See

PR #196, opened 2019-03-21
add custom issuer checker

ti:
Something times, the wildchard domain or subdoamin need a custom  issuer checker
Like: 

https://firebase.google.com/docs/auth/admin/verify-id-tokens#verify_id_tokens_using_a_third-party_jwt_library

| ID Token Payload Claims |                     |                                                              |
| :---------------------- | ------------------- | ------------------------------------------------------------ |
| `aud`                   | Audience            | Must be your Firebase project ID, the unique identifier for your Firebase project, which can be found in the URL of that project's console. |
| `iss`                   | Issuer              | Must be `"https://securetoken.google.com/<projectId>"`, where `<projectId>` is the same project ID used for `aud` above. |
| `sub`                   | Subject             | Must be a non-empty string and must be the `uid` of the user or device. |
| `auth_time`             | Authentication time | Must be in the past. The time when the user authenticated.   |

> We do not recommend turning off Checker.

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -87,6 +87,8 @@ type Config struct {
 	// If true, token expiry is not checked.
 	SkipExpiryCheck bool
 
+	IssuerChecker func(issuer string) bool
+
 	// Time function to check Token expiry. Defaults to time.Now
 	Now func() time.Time
 }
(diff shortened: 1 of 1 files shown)


ti:
@ericchiang, some times we really need a custom checker, and skip method is foo.

https://github.com/coreos/go-oidc/pull/195

ericchiang:
If your project ID is "foo-bar" why can't you use the issuer `https://securetoken.google.com/foo-bar`? Why does this need to be a dynamic check?

ti:
@ericchiang.

I my case, the resources server behind oidc, must check check multiple content.

my server is a serverless or BaaS service for small business, the resources server need a wildchard to check the domain.


the  openid-configuration is:

```json
{
      "issuer": "https://foo-bar.xxxxx.cloud",
      "authorization_endpoint": "https://foo-bar.xxxxx.cloud/__/authorize",
       "token_endpoint": "https://foo-bar.xxxxx.cloud/v1/auth/token",
       "userinfo_endpoint": "https://foo-bar.xxxxx.cloud/v1/user/me",
       "revocation_endpoint": "https://foo-bar.xxxxx.cloud/v1/auth/revoke",
      "jwks_uri": "https://foo-bar.xxxxx.cloud/v1/auth/certs",
       ....
}
```

> the  `foo-bar` is the project id


for the project user the current go-oidc lib is ok, but for my `addtional resources server`, we trust all `*.xxxxx.cloud`, because our back-end resources will be namespaced according to different `project_id`.
and the `project_id` is dynamic, if every project id has a Independent openid， tens of thousands of project ids may not work.

and, for more, the `token_endpoint` and `userinfo_endpoint` also have project in the domain. we can not use same  openid-configuration for every project

verify.go
```diff
@@ -87,6 +87,8 @@ type Config struct {
 	// If true, token expiry is not checked.
 	SkipExpiryCheck bool
 
+	IssuerChecker func(issuer string) bool
```
ericchiang, line 90: Instead of this, let's do `SkipIssuerCheck` and add a good doc string:

```
// SkipIssuerCheck is intended for specialized cases where the the caller wishes to defer
// issuer validation. When enabled, callers MUST independently verify the Token's Issuer
// is a known good value.
//
// Mismatched issuers often indicate client mis-configuration. If mismatches are unexpected,
// evaluate if the provided issuer URL is incorrect instead of enabling this option.
SkipIssuerCheck bool
```

ericchiang (commented):
This seems fine since it's not an argument to `NewProvider`. It's similar to SkipClientIDCheck which is equally niche.

ericchiang:
bump

bshi:
@ericchiang I've need of SkipIssuerCheck - would you be willing to take a pull request implementing your suggestion?

PR #202, opened 2019-06-26
*: add a SkipIssuerCheck option to the validation config

ericchiang:
For special cases, it's sometimes fine to ignore the issuer check on an
ID token, similar to SkipClientIDCheck. Include appropriate warnings.

I don't expect to add a similar option for NewProvider. Every time that's
been asked for, it's indicated a client misconfiguration.

Closes #196
Closes #195 

cc @bshi


bshi:
LGTM

Much appreciated!

PR #205, opened 2019-07-10
Support raw UserInfo response bodies

joshua-rutherford:
This is a very simplistic approach to allowing users to retrieve signed and/or encrypted responses from the UserInfo endpoint.  I realize this is a very simplistic that leaves much up to the user who actually wants to use this, but it's an approach.


joshua-rutherford:
Closing in response to @ericchiang's workable solution proposed in #206.

Issue #203, opened 2019-07-03
Ignore issuer trailing slash mismatch in gooidc.NewProvider


aeneasr:
Providing issuer url `http://foo.com` to `gooidc.NewProvider` results in an issue if the OpenID Connect Discovery Endpoint includes a trailing slash in it's issuer value: `http://foo.com/`.

The spec does not state if the issuer value should contain a trailing slash or not. The URLs are - at least at the root level - equal.

Because some issuer use trailing slashes and others do not, I would like to propose (and happy to pull up a PR) to ignore trailing slash mismatches between what's passed to `gooidc.NewProvider` and what is returned by `issuer`.

I **do not** propose to introduce this logic to the actual verification of the ID Token `iss`.

ericchiang:
> The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve the configuration information. This MUST also be identical to the iss Claim value in ID Tokens issued from this Issuer.

https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation

I read that as "the strings must match".

Beyond the spec, the issuer is treated as a string throughout this package. We'd consider a token that returned `http://foo.com` when the metadata specified `http://foo.com/` invalid. I personally like the lack of fuzzy matching, it's way simpler to reason about.

Which provider does this and how big of an issue is it? You try to create an issuer once, get a mismatch error, then fix it. Can that provider improve its docs or could we surface a better error message?

aeneasr:
> I read that as "the strings must match".

I actually missed that part and agree. Therefore I don't think that there's really anything this library can do beyond what it's doing right now. [ory/hydra](https://github.com/ory/hydra) is always appending a trailing slash to the issuer value, which causes issues in cases where developers switch an existing OP with it and where the issuer value changes from no trailing slash to trailing slash. However, I don't think that's something that can be fixed here due to the language in the spec.

Maybe it would be possible to improve the error message as it is quite easy to overlook the trailing slash.

ericchiang:
Going to close this for now as it looks like this is an issue with hydra. If you have a suggestion for improving the error message feel free to send a PR :)

PR #209, opened 2019-08-09
Fix unexpected behavior of comparison an issuer URLs

binlab:
Fixed unexpected behavior observed during validation `issuer` URL from configuration and received from a `provider`. As a different provider can return different format of URL condition couldn't determine exactly match of URLs.
An error was noticed when using **Vault** which uses this library for **OIDC Auth Backend**.
Issue of error: https://github.com/hashicorp/vault/issues/7285
Here is an example and how to reproduce on different providers :

**Google:**

`https://accounts.google.com/.well-known/openid-configuration`
```
{
 "issuer": "https://accounts.google.com",
 "authorization_endpoint": "https://accounts.google.com/o/oauth2/v2/auth",
 "token_endpoint": "https://oauth2.googleapis.com/token",
 ...
}
```

**Auth0:**

`https://mycompany.auth0.com/.well-known/openid-configuration`
```
{
 "issuer":"https://mycompany.auth0.com/",
 "authorization_endpoint":"https://mycompany.auth0.com/authorize",
 "token_endpoint":"https://mycompany.auth0.com/oauth/token",
 ...
}
```

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -120,7 +120,7 @@ func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
 		return nil, fmt.Errorf("oidc: failed to decode provider discovery object: %v", err)
 	}
 
-	if p.Issuer != issuer {
+	if strings.TrimSuffix(p.Issuer, "/") != strings.TrimSuffix(issuer, "/") {
 		return nil, fmt.Errorf("oidc: issuer did not match the issuer returned by provider, expected %q got %q", issuer, p.Issuer)
 	}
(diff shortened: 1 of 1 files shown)


ericchiang:
Sorry, this is a dup of https://github.com/coreos/go-oidc/issues/203

Issuers are exact strings. Even if you do fuzzy matching here, you have to consider what the ID Token provides as well. I'd rather see this resolved with docs. For your second example, you can already provide `https://mycompany.auth0.com/`

PR #210, opened 2019-08-15
Honour nbf claim if present in ID token

anderseknert:
As [promised](https://github.com/kubernetes/kubernetes/pull/81413#issuecomment-521526423) :) Not new to the domain but very new to Go programming, so bear with me.

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -270,9 +270,20 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		if v.config.Now != nil {
 			now = v.config.Now
 		}
+		nowTime := now()
 
-		if t.Expiry.Before(now()) {
+		if t.Expiry.Before(nowTime) {
 			return nil, fmt.Errorf("oidc: token is expired (Token Expiry: %v)", t.Expiry)
+		}
+
+		// If nbf claim is provided in token, ensure that it is indeed in the past.
+		if token.NotBefore != nil {
+			nbfTime := time.Time(*token.NotBefore)
+			leeway := 1 * time.Minute
+
+			if nowTime.Add(leeway).Before(nbfTime) {
+				return nil, fmt.Errorf("oidc: current time %v before the nbf (not before) time: %v", nowTime, nbfTime)
+			}
 		}
 	}
 
--- a/verify_test.go
+++ b/verify_test.go
@@ -107,6 +107,34 @@ func TestVerify(t *testing.T) {
 			},
 			signKey: newRSAKey(t),
 		},
+		{
+			name: "nbf in future",
+			idToken: `{"iss":"https://foo","nbf":` + strconv.FormatInt(time.Now().Add(time.Hour).Unix(), 10) +
+				`,"exp":` + strconv.FormatInt(time.Now().Add(time.Hour).Unix(), 10) + `}`,
+			config: Config{
+				SkipClientIDCheck: true,
+			},
+			signKey: newRSAKey(t),
+			wantErr: true,
+		},
+		{
+			name: "nbf in past",
+			idToken: `{"iss":"https://foo","nbf":` + strconv.FormatInt(time.Now().Add(-time.Hour).Unix(), 10) +
+				`,"exp":` + strconv.FormatInt(time.Now().Add(time.Hour).Unix(), 10) + `}`,
+			config: Config{
+				SkipClientIDCheck: true,
+			},
+			signKey: newRSAKey(t),
+		},
+		{
+			name: "nbf in future within clock skew tolerance",
+			idToken: `{"iss":"https://foo","nbf":` + strconv.FormatInt(time.Now().Add(30*time.Second).Unix(), 10) +
+				`,"exp":` + strconv.FormatInt(time.Now().Add(time.Hour).Unix(), 10) + `}`,
+			config: Config{
+				SkipClientIDCheck: true,
+			},
+			signKey: newRSAKey(t),
+		},
 	}
 	for _, test := range tests {
 		t.Run(test.name, test.run)
--- a/oidc.go
+++ b/oidc.go
@@ -321,6 +321,7 @@ type idToken struct {
 	Audience     audience               `json:"aud"`
 	Expiry       jsonTime               `json:"exp"`
 	IssuedAt     jsonTime               `json:"iat"`
+	NotBefore    *jsonTime              `json:"nbf"`
 	Nonce        string                 `json:"nonce"`
 	AtHash       string                 `json:"at_hash"`
 	ClaimNames   map[string]string      `json:"_claim_names"`


verify.go
```diff
@@ -276,6 +279,32 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
 
+	// If nbf claim is provided in token, ensure that it is indeed in the past.
+	if !v.config.SkipNbfCheck {
+		var claimsMap map[string]interface{}
+		err := json.Unmarshal(t.claims, &claimsMap)
+		if err != nil {
+			return nil, fmt.Errorf("oidc: failed to unmarshal claims: %v", err)
+		}
+		if nbfRaw := claimsMap["nbf"]; nbfRaw != nil {
+			nbf, ok := nbfRaw.(float64)
```
mikedanese: Add NotBeforeField to IDToken starting from here:

https://github.com/coreos/go-oidc/blob/8ae1da518bd4d9d5a5909090a184af30f336436d/oidc.go#L323

since it's a public claim.

verify.go
```diff
@@ -276,6 +279,32 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		}
 	}
 
+	// If nbf claim is provided in token, ensure that it is indeed in the past.
+	if !v.config.SkipNbfCheck {
```
mikedanese: @ericchiang is this worth a seperate check or should we fold this into SkipExpiryCheck?

verify.go
```diff
@@ -276,6 +279,32 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
+			if !ok {
+				return nil, fmt.Errorf("oidc: invalid nbf claim, not a number: %v", err)
+			}
+			nbfTime := time.Unix(int64(nbf), 0)
+
+			now := time.Now
+			if v.config.Now != nil {
+				now = v.config.Now
+			}
+			nowTime := now()
```
mikedanese: If we go with SkipExpiryCheck, move:

https://github.com/coreos/go-oidc/blob/274971e2c94cd6fb17614979e2899edbc81af146/verify.go#L269-L272

Up one block and reuse now so both checks work of the same time.

verify.go
```diff
@@ -276,6 +279,32 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
+
+			if nowTime.Before(nbfTime) {
```
mikedanese: Add a minute of leeway here to forgive some clock skew. `nowTime.Add(1*time.Minute).Before(nbfTime)`.

verify_test.go
```diff
@@ -107,6 +107,16 @@ func TestVerify(t *testing.T) {
 			},
 			signKey: newRSAKey(t),
 		},
+		{
+			name:    "nbf",
```
mikedanese: we need a positive and negative test.

verify.go
anderseknert: Thanks, did so and added a test case for that.

verify_test.go
anderseknert: Fixed.

verify.go
anderseknert: My rationale was that the nbf claim is not [ID Token specific](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) but a more ["common" claim](https://tools.ietf.org/html/rfc7519#section-4.1.5) which none of the other ones listed so far is - perhaps there is some value of keeping a distinction between the two? If not I'll add a NotBeforeField as you suggest.

verify.go
ericchiang: Yeah, just group it under SkipExpiryCheck. No need to add another public flag.

verify.go
ericchiang: That struct isn't exported, it's solely used for this method and validation :) Feel free to add any fields to it.

You can also use a pointer to determine presence

```go
type idToken struct {
    IssuedAt  jsonTime  `json:"iat"`
    NotBefore *jsonTime `json:"nbf"`
}

// ...

var t idToken
// ... unmarshal

if t.NotBefore != nil {
    // "nbf" is present. Evaluate it
}
```

anderseknert: Nice one! Thanks 👍 Check looking so much simpler now.

verify.go
anderseknert: Done - simpler for sure, though the tests got a little more verbose as I now had to include an `exp` in all of the `nbf` test claims or they'd fail.

anderseknert:
Thanks for all the great feedback guys! I think I've addressed your points - let me know if not :)

verify.go
```diff
@@ -274,6 +274,17 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
 		}
+
+		// If nbf claim is provided in token, ensure that it is indeed in the past.
+		if token.NotBefore != nil {
+			nbfTime := time.Time(*token.NotBefore)
+			nowTime := now()
+			leeway := 1 * time.Minute
+
+			if nowTime.Add(leeway).Before(nbfTime) {
+				return nil, fmt.Errorf("oidc: current time %d before the nbf (not before) time: %d", nowTime.Unix(), nbfTime.Unix())
```
ericchiang: nit: for consistency with the token expiry above, just used %v for the time formatting

```
return nil, fmt.Errorf("oidc: current time %v before nbf (not before) time: %v", nowTime, nbfTime)
```

verify.go
```diff
@@ -274,6 +274,17 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		if t.Expiry.Before(now()) {
 			return nil, fmt.Errorf("oidc: token is expired (Token Expiry: %v)", t.Expiry)
 		}
+
+		// If nbf claim is provided in token, ensure that it is indeed in the past.
+		if token.NotBefore != nil {
+			nbfTime := time.Time(*token.NotBefore)
+			nowTime := now()
```
ericchiang: Another nit: make sure `nowTime` is the same for the Expiry and NotBefore check

```
nowTime := now()
if t.Expiry.Before(nowTime) {
    // handle error
}
if token.NotBefore != nil {
    nbfTime := time.Time(*token.NotBefore)
    leeay := 1 * time.Minute
    if nowTime.Add(leeway).Before(nbfTime) {
        // handle error.
    }
}
```

ericchiang (approved):
just a couple nits

please squash your commits into a single commit :)

verify.go
anderseknert: Yeah I think there's a point to be made for printing the values in the same form as they come in the token, but consistency is key of course. Fixed.

anderseknert:
Thanks for guiding me through this one @ericchiang @mikedanese . Much appreciated 👍

PR #217, opened 2019-10-08
Allow query params in issuer URL

ifross89:
Some OpenID providers (e.g. Azure Active Directory B2C) have query parameters in their issuer URL.

Allow such issuers to be used by parsing the issuer URL to form the "well known" URL by manipulating the URL path.


ericchiang:
Please open an issue describing the overall problem with links to the azure docs. It's generally better practice to discuss background before sending a PR :)

Issue #219, opened 2019-10-17
The issuer should not be known by the client.  Only authority is needed.


ghstahl:
```
provider, err := oidc.NewProvider(ctx, "https://localhost:6001")
....
if p.Issuer != issuer {
		return nil, fmt.Errorf("oidc: issuer did not match the issuer returned by provider, expected %q got %q", issuer, p.Issuer)
	}
```

1. The spec does not state that the "Authority" == issuer, which this code assumes.  
2. The issuer can point to anything and does not need to be in the same domain.  i.e. Authority = https://a.b.com, issuer = https://a.cat.com 

The issuer returned by 

> /.well-known/openid-configuration

 needs only to match the iss in the JWT tokens that follow.    The only knowledge the client should have is "authority".  

the call should be;
```
authority := "https://localhost:6001"
issuer := "<you will be told this by openid-configuration>"
provider, err := oidc.NewProvider(ctx,authority)  
```

Using this code will break in the field if an IDP like google decides to change the issuer, which by spec they can.

ericchiang:
Dup of the following issues?

https://github.com/coreos/go-oidc/issues/213
https://github.com/coreos/go-oidc/issues/204
https://github.com/coreos/go-oidc/issues/203
https://github.com/coreos/go-oidc/issues/183
https://github.com/coreos/go-oidc/issues/159

I don't really see anything new in this issue. The spec clearly states that when doing http discovery, the "iss" MUST identically match the URL provided.

What provider or setup are you having issues with?

mikedanese:
To prevent impersonation attack:

> RPs MUST ensure that the Issuer URL they are using for the Configuration Request exactly matches the value of the issuer Claim in the OP Metadata document received by the RP and that this also exactly matches the iss Claim value in ID Tokens that are supposed to be from that Issuer.

https://openid.net/specs/openid-connect-discovery-1_0.html#Impersonation

ghstahl:
**To prevent impersonation attack**: I cover this on my exposed APIS, I don't need to cover it when a client obtains an access_token.

The spec talks about Validation, and not discovery rules.

4.3.  OpenID Provider Configuration Validation
If any of the validation procedures defined in this specification fail, any operations requiring the information that failed to correctly validate MUST be aborted and the information that failed to validate MUST NOT be used.

The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve the configuration information. This MUST also be identical to the iss Claim value in ID Tokens issued from this Issuer.

[...]

PR #220, opened 2019-10-17
https://github.com/coreos/go-oidc/issues/219

ghstahl:
issuer is not to be known by the client, it is given by /.well-known/openid-configuration

https://github.com/coreos/go-oidc/issues/219


ericchiang:
Per discussion in #219, going to close this out.

PR #226, opened 2019-12-20
Added logout functionality through provider

dlouwers:
Replacement for #190 since I need to create more pull requests now

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -69,6 +70,7 @@ type Provider struct {
 	authURL     string
 	tokenURL    string
 	userInfoURL string
+	logoutURL   string
 
 	// Raw claims returned by the server.
 	rawClaims []byte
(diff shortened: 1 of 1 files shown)


dlouwers:
Seems to be an issue with the Go 1,9 build. Unclear what since it is the same PR that passed but from a different branch: ```The command "go get -v -t github.com/coreos/go-oidc/..." failed and exited with 2 during .```

SachinVarghese:
@dlouwers Hey any updates on this? We are working on a similar use case where we need a redirect to a logout URL.

brocaar:
@dlouwers is there any update on this? It looks like the Go 1.9 error causing the build to fail is:

> ../../../gopkg.in/square/go-jose.v2/encoding.go:58:9: **undefined: strings.Builder**

I'm looking forward to use this feature :)

StiviiK:
> @dlouwers is there any update on this? It looks like the Go 1.9 error causing the build to fail is:
> 
> > ../../../gopkg.in/square/go-jose.v2/encoding.go:58:9: **undefined: strings.Builder**
> 
> I'm looking forward to use this feature :)

You need to use Go 1.10+ as this was added with Go 1.10
https://stackoverflow.com/questions/48978414/golang-strings-builder-type-undefined

brocaar:
Would re-running the tests be the solution (and resolve the merge conflict)? Looks like Travis CI config has been updated to Go 1.12 and Go 1.13 since the test failed: https://github.com/coreos/go-oidc/blob/v2/.travis.yml#L4.

mitar:
+1 for getting this in.

mitar:
+1 for getting this in.

mitar:
For anyone else going around this, you can currently do something like:

```go
	provider, err := oidc.NewProvider(ctx, issuer)
	if err != nil {
		return nil, err
	}

	var claims struct {
		EndSessionURL string `json:"end_session_endpoint"`
	}
	err = provider.Claims(&claims)
	if err != nil {
		return nil, err
	}
	// ... claims.EndSessionURL is now end session URL to use
```

You have to construct final URL yourself, but it gets you moving.

ghost (approved):

ABGEO:
Hello,

Any updates on this?

jschlechtengrok:
+1 This would be a welcome addition to the library!

PR #227, opened 2020-01-23
*: pass algorithms a provider supports to the verifier

### LINKED ISSUE #225: Supported Signing Algorithms: Add Support to the Provider struct for parsing the supported signing algorithms
Hi all!
We have been using the go-oidc library to build an [oidc-authservice](https://github.com/arrikto/oidc-authservice) for the Kubeflow Istio Gateway and a user stumbled on an unsupported signing algorithm error while using a different from RS256 signing algorithm. (https://github.com/kubeflow/kubeflow/issues/4550). 

The code we use for getting a verifier is: 
```go
verifier := s.provider.Verifier(&oidc.Config{ClientID: s.oauth2Config.ClientID})
```

I would expect that the provider would save the supported signing algorithms for idtoken and userinfo in its internal struct. Those are give in the standard fields `id_token_signing_alg_values_supported` and `userinfo_signing_alg_values_supported`.

I think this is something we would want to support, otherwise I have to manually scrape the well-known endpoint for that info.
If you're ok with it, I can put the cycles to prepare a PR for this.

/cc @ericchiang

ericchiang:
If a provider advertises support for algorithms like ES256 in its
discovery, pass those algorithms to the verifier. Added a whitelist to
ensure weird algorithms like "none" are thrown out.

Fixes #225

### DIFF
--- a/oidc_test.go
+++ b/oidc_test.go
@@ -1,7 +1,13 @@
 package oidc
 
 import (
+	"context"
 	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


oidc_test.go
```diff
@@ -93,3 +99,198 @@ func TestAccessTokenVerification(t *testing.T) {
...
+				w.Header().Set("Content-Type", "application/json")
+				io.WriteString(w, strings.ReplaceAll(test.data, "ISSUER", issuer))
+			}
+			s := httptest.NewServer(http.HandlerFunc(hf))
+			defer s.Close()
+			issuer = s.URL
+			if test.trailingSlash {
+				issuer += "/"
+			}
+			p, err := NewProvider(ctx, issuer)
```
mikedanese, line 270: A few newlines would make this read a little better.

mikedanese:
lgtm

oidc_test.go
ericchiang: done

PR #228, opened 2020-01-23
*: add a go.mod file

### LINKED ISSUE #218: Go module support
Any chance we could have go module support?

ericchiang:
Generated using the following commands:

    go mod init github.com/coreos/go-oidc
    go mod tidy

Fixes #218

NOTE: this is expected to fail due to travis issues until #227 is merged and this is rebased.


mikedanese:
Should we bump these?

https://github.com/coreos/go-oidc/blob/2be1c5b8a260760503f66dc0996e102b683b3ac3/.travis.yml#L4-L5

mikedanese:
> Each major Go release is supported until there are two newer major releases.

https://golang.org/doc/devel/release.html#policy

1.13, 1.12?

ericchiang:
Bumped in #227 :)

ericchiang:
I'll pull it out into this PR

pjediny:
Should not this be /v2 or better /v3 module?

ericchiang:
Happy to do that for later breaking changes, but I'd rather not force users to change the import path today.

pjediny:
Would you not need to change it anyway from `gopkg.in/coreos/go-oidc.v2` to `github.com/coreos/go-oidc`? I'm not sure how the go would work with this module, as they mandate the semver major change to be reflected by import paths.  But I don't know how this works with `gopkg.in`, having special exception in module versioning.

ericchiang:
This package doesn't use gopkg.in.

Though it does look like some people import it this way... https://godoc.org/gopkg.in/coreos/go-oidc.v2?importers

I'll open bugs against those importers.

pjediny:
And my private project too :-), hence the confusion.
If I can suggest, maybe it should be documented in the `README.md` in a similar way it is done in `github.com/square/go-jose`.

ericchiang:
This is documented in the godoc link, examples, and now the go.mod. I think that should be good :)

PR #229, opened 2020-01-25
Fix issues with long-held reference to context passed to NewProvider

### LINKED ISSUE #214: updateKeys for remoteKeySet uses wrong context
Im trying to verify token with Verifier created by provider. 

```
verifier := provider.Verifier(&oidc.Config{ClientID: myClientID})
rawIDToken, ok := oauth2Token.Extra("id_token").(string)
ctx,_ = context.WithTimeout(context.Background(), time.Second*10)
idToken, err:= verifier.Verify(ctx, rawIDToken)
```
less than in a second i got the error saying `context deadline exceeded`
Trying to figure out what happened I changed `ctx` to `context.Background()`
but issue still persisted.

Digging into Verify code i find out that keys are actually fetched by function 
`func (r *remoteKeySet)keysFromRemote(ctx context.Context)([]jose.JSONWebKey, error)`
and inside goroutine there is call `r.updateKeys()` which does not receive a context and inside the `updateKeys` actually is used another context `r.ctx` which one assigned by Provider.

Provided is created at start of the application, and due http request for openid-configuration Im not updating it often. and due same http request I set a context for provider with a deadline for 10 seconds, just in case if there is no connection to provider.

And what happens is Verify is actually would work during first 10 seconds after provider has been created, but not considering context passed to `Verify`

misberner:
Instead of using a context to bound key fetch requests, allow the user to pass
a custom HTTP client to control timeout behavior. There are no behavioral
changes for the now-deprecated NewRemoteKeySet function.

Change NewProvider to only use the passed context for the discovery HTTP
request and for extracting an HTTP client to be used in the remote key set.

Enforce a default HTTP request timeout of 30 seconds if the client for remote
key fetches does not specify a timeout.

Fixes #214

NOTE: See #214 for a discussion of the issue. This change aims at being minimally disruptive (of course some behavioral change is inevitable) without taking any features that specifying a context entails away. This motivates the choice to pass an `*http.Client` to `NewRemoteKeySetWithClient` in favor of a simpler option such as a simple `timeout time.Duration`.

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/pquerna/cachecontrol"
+	"golang.org/x/oauth2"
 	jose "gopkg.in/square/go-jose.v2"
 )
 
@@ -23,6 +24,21 @@ import (
 // updated.
 const keysExpiryDelta = 30 * time.Second
 
+// NewRemoteKeySetWithClient returns a KeySet that can validate JSON web tokens by using
+// GETs to fetch JSON web token sets hosted at a remote URL. This is automatically
+// used by NewProvider using the URLs returned by OpenID Connect discovery, but is
+// exposed for providers that don't support discovery or to prevent round trips to the
+// discovery URL.
+//
+// The returned KeySet is a long lived verifier that caches keys based on cache-control
+// headers. Reuse a common remote key set instead of creating new ones as needed.
+//
+// The client parameter controls the behavior of the fetch requests, such as the request
+// timeout. A value of nil causes http.DefaultClient to be used.
+func NewRemoteKeySetWithClient(client *http.Client, jwksURL string) KeySet {
+	return newRemoteKeySet(context.Background(), client, jwksURL, time.Now)
+}
+
 // NewRemoteKeySet returns a KeySet that can validate JSON web tokens by using HTTP
 // GETs to fetch JSON web token sets hosted at a remote URL. This is automatically
 // used by NewProvider using the URLs returned by OpenID Connect discovery, but is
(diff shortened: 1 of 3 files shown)


ericchiang:
I think this could actually be fixed by adding a method to construct a Verifier with a context:

```go
func (p *Provider) VerifierContext(ctx context.Context, config *Config) (*IDTokenVerifier) {}
```

...or NewVerifier if that's a better name. So if you did:

```go
p, err := oidc.NewProvider(pCtx, issuerURL)
if err != nil {
    // ...
}
v, err := p.VerifierContext(vCtx, config)
```

Then `NewProvider` would be bounded by `pCtx`, and the created verifier would use `vCtx`.

misberner:
Hmm, this would solve the problem of not being able to effectively bound `NewProvider` via the context, but it doesn't really address the anti-pattern of having a long-lived context in the struct - I can't think of a reason why I would pass anything other than `context.Background()` to `VerifierContext`, given that `(*IDTokenVerifier).Verify` already takes a context param. It also doesn't make controlling the timeout for the background fetch any easier.

ericchiang:
There are two properties that need to hold:

* IDTokenVerifier needs to be able to be shut down, either by canceling a context or by calling a Close method.
* Canceling a context associated with a Verify() shouldn't impact the background routine refreshing the keys.

I know the long running context isn't idiomatic, but we can't change the behavior of the package for existing users without going to v3. e.g. this code needs to have the same behavior no matter what for v3:

```
ctx, cancel := context.WithCancel(context.Background())
p, err := oidc.NewProvider(ctx, issuerURL)
if err != nil {
    // ...
}
v := p.Verifier(config)
cancel()
```

ericchiang:
Fixed by #262 which will be on the v3 branch.

PR #237, opened 2020-03-01
Provider: added end session support

dzen-it:
This is a popular feature that is useful when working with the library.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -65,11 +66,12 @@ func doRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
 
 // Provider represents an OpenID Connect server's configuration.
 type Provider struct {
-	issuer      string
-	authURL     string
-	tokenURL    string
-	userInfoURL string
-	algorithms  []string
+	issuer        string
+	authURL       string
+	tokenURL      string
+	endSessionURL string
+	userInfoURL   string
+	algorithms    []string
 
 	// Raw claims returned by the server.
 	rawClaims []byte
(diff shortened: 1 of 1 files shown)


oidc.go
```diff
@@ -238,6 +242,73 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
...
 
+// EndSessionOptions is optional parameters for the Initiated Logout by https://openid.net/specs/openid-connect-session-1_0.html#RPLogout
+type EndSessionOptions struct {
+	IDTokenHint string
+	RedirectURL string
+	State       string
+}
+
+// EndSession uses the token source to query the provider's end session endpoint.
+func (p *Provider) EndSession(ctx context.Context, tokenSource oauth2.TokenSource, options EndSessionOptions) error {
```
ericchiang, line 253: nit: rename "options" to "o"

oidc.go
```diff
@@ -238,6 +242,73 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
...
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	return nil
+}
+
+func makeEndSessionEndpoint(endSessionURL string, options EndSessionOptions) (endURL string, err error) {
```
ericchiang, line 287: nit: rename "options" to "o"
nit: rename "makeEndSessionEndpoint" to "endSessionEndpoint"

Also please use shorter variable names

```go
	if endSessionURL == "" {
		return "", errors.New("oidc: end session endpoint is not supported by this provider")
	}

	u, err := url.Parse(endSessionURL)
	if err != nil {
		return "", fmt.Errorf("oidc: wrong format of end session endpoint %s: %v", p.endSessionURL, err)
	}
	q := u.Query()
	if o.IDTokenHint != "" {
		q.Set("id_token_hint", o.IDTokenHint)
	}
	if o.RedirectURL != "" {
		q.Set("post_logout_redirect_uri", o.RedirectURL)
	}
	if o.State != "" {
		q.Set("state", o.State)
	}
	u.RawQuery = q.Encode()
	return u.String(), nil
```

ericchiang: Also, please add a test for this method.

oidc.go
```diff
@@ -238,6 +242,73 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
...
+	}
+
+	token, err := tokenSource.Token()
+	if err != nil {
+		return fmt.Errorf("oidc: get access token: %v", err)
+	}
+
+	token.SetAuthHeader(req)
+
+	resp, err := doRequest(ctx, req)
```
ericchiang, line 271: I'm not sure it's correct for the library to make this request. The spec specifically calls out that this should be initiated by the browser, not the server

```
An RP can notify the OP that the End-User has logged out of the site, and might want to log out of the OP as well. In this case, the RP, after having logged the End-User out of the RP, redirects the End-User's User Agent to the OP's logout endpoint URL. This URL is normally obtained via the end_session_endpoint element of the OP's Discovery response, or may be learned via other mechanisms.
```
https://openid.net/specs/openid-connect-session-1_0-17.html#RPLogout

ericchiang:
Also when you mention "This is a popular feature that is useful when working with the library." do you have a list of providers that actually implement this feature? :)

dzen-it:
I looked carefully at the implementation of the top providers, and  you are right, the "end session" should be cooked on the client side. Thank you for your time.

PR #239, opened 2020-03-13
Ignore trailing slash when checking the issuer from the OpenID discovery URL

### LINKED ISSUE #238: Issuer check in oidc.NewProvider - strict issuer checking seems unnecessary
## What

When a new provider instance is created, it goes to the discovery url (`/.well-known/openid-configuration`) and checks whether the issuer parameter matches exactly the issuer variable value.

NewProvider throws an error if you've included a trailing slash at the end of the issuer, but the discovery url returns the value without it and vice-versa. It seems to me this strict check has no useful purpose and makes misconfiguration issues easier.

I propose to ignore trailing slash differences in NewProvider, to have a better user experience.

As a previous issue about this was closed (https://github.com/coreos/go-oidc/issues/66) I wanted to ask first whether you would be open about receiving a change PR for this.

bandesz:
## What

Previously the NewProvider function threw an error if the passed `issuer`
variable and the `issuer` parameter from the discovery URL didn't exactly
match.
This should not happen if the only difference is a trailing slash.

This fixes https://github.com/coreos/go-oidc/issues/238

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -111,7 +111,8 @@ var supportedAlgorithms = map[string]bool{
 // The issuer is the URL identifier for the service. For example: "https://accounts.google.com"
 // or "https://login.salesforce.com".
 func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
-	wellKnown := strings.TrimSuffix(issuer, "/") + "/.well-known/openid-configuration"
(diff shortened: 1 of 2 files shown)


ericchiang:
Closing per https://github.com/coreos/go-oidc/issues/238#issuecomment-603556858

Issue #232, opened 2020-01-27
Changes for v3


ericchiang:
Since we need to modify the import path to make a go.mod work (#230), it might be a good chance to fix some APIs and call it v3.

Some changes that seem uncontroversial:

* Change import to github.com/coreos/go-oidc/v3/oidc
* Remove import of github.com/pquerna/cachecontrol (#127)
* Separate NewProvider context and Verifier context (#214)

Changes that are more intrusive:

* Don't use context keys to configure http.Clients? This pattern is popular in Google packages, but is generally discouraged (https://github.com/golang/go/issues/18284). Would require a lot of API changes. 
* Remove golang.org/x/oauth2 import? If you're just verifying ID Tokens, you generally don't need it. Would require changing the following signatures:

```
func Nonce(nonce string) oauth2.AuthCodeOption
func (p *Provider) Endpoint() oauth2.Endpoint
func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource) (*UserInfo, error)
```

ericchiang:
Am going to move forward with this so we can add a go.mod file. I think I'll forgo the context and oauth2 import changes.

ericchiang:
Hmm, some issues with the v3 path. Investigating

```
go get github.com/coreos/go-oidc/oidc@v3.0.0-alpha.1: github.com/coreos/go-oidc@v3.0.0-alpha.1: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v3
```

Edit: figured it out

```
go get github.com/coreos/go-oidc/v3/oidc@v3.0.0-alpha.1
```

ericchiang:
Okay I think we're at the point we can tag v3. Going to send a README update then call it done.

ericchiang:
README PR sent (#232). Once that's in, I'll tag v3.0.0 and change the default branch to v3.

ericchiang:
Tagged https://github.com/coreos/go-oidc/releases/tag/v3.0.0

PR #247, opened 2020-05-05
*: add a go.mod file and migrate to v3 package

ericchiang:
Adds a go.mod file and moves the package to the "oidc" directory.
Everything in this PR is just a move except:

* Changes to the test script for go.mods.
* Moves oidc/jose.go comments to above the const block to satisfy golint.
* EDIT: TravisCI was taking a while so I also moved to GitHub Actions. No idea how
to trigger it without merging.

Updates #232


.github/workflows/test.yaml
```diff
@@ -0,0 +1,22 @@
+name: test
+on: 
+  pull_request:
+    branches:
+      - v3
```
mfridman: What about enabling actions for any PRs and the default branch: `v3` ?

```suggestion
on:
  push:
    branches:
      - v3
  pull_request:
```

go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/coreos/go-oidc/v3
+
+go 1.14
+
+require (
+	github.com/coreos/go-oidc v2.2.1+incompatible
```
mfridman: One suggestion is to run:

```
go mod tidy && go mod verify
```

Which will remove v2 references in go.mod and go.sum

.github/workflows/test.yaml
ericchiang: sgtm, added

go.mod
ericchiang: woops, yep done

PR #249, opened 2020-05-26
handle UserInfo.email_verified as a string

calebdoxsey:
We were using this library to interface with AWS Cognito and received this error message:

```json: cannot unmarshal string into Go struct field UserInfo.email_verified of type bool```

So apparently they've decided to return the `email_verified` field as a string instead of a bool. 

These changes make the code more tolerant of the type of data so that it can handle both cases.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ericchiang** [comment]
email_verified should explicitly be a bool according to the spec:

https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

Out of curiosity, why are you using UserInfo? You can get this information from the IDToken without an additional round trip:

```go
rawIDToken, ok := oauth2Token.Extra("id_token").(string)
if !ok {
    // handle error
}
idToken, err := idTokenVerifier.Verify(rawIDToken)
if err != nil {
	// handle error
}
var claims struct {
	Email         string `json:"email"`
	EmailVerified string `json:"email_verified"`
}
if err := idToken.Claims(&claims); err != nil {
	// handle error
}
```

calebdoxsey:
@ericchiang thanks for getting back to me so quickly. I believe we use the additional data returned by UserInfo for additional policy routing and enforcement. (in [Pomerium](https://github.com/pomerium/pomerium))

It's unfortunate that AWS Cognito has chosen not to obey the spec. I suppose it makes sense to keep the library clean of workarounds like this. I can solve this in my own code if that seems like the better option.

calebdoxsey:
I came up with a workaround by modifying the response from AWS cognito before sending it to the OIDC code by using a custom HTTP client stored in the context.

A future improvement to this library might be some mechanism to make it easier to make these kinds of transformations possible, but I think my workaround is sufficient for now. 

Thanks again for the feedback. I'll close this PR.

smartinez-yatiribio:
Hi @calebdoxsey . I am having the same problem. I get the error:
`OAuth2 handling error trying to obtaining user record: oidc: failed to decode userinfo: json: cannot unmarshal string into Go struct field UserInfo.email_verified of type bool`
I am trying to solve it by adding a Lambda function in my Cognito user pool triggered at _**Pre token generation Lambda**_
In that lambda function, I try to remove or change the attribute **`email_verified`** but it doesn't work...I get always the same error.
I am wondering if you could explain more in detail what was your workaround about this.

calebdoxsey:
@smartinez-yatiribio the workaround can be seen here: https://github.com/pomerium/pomerium/blob/0b0fba06b3374557ed7427d165190570ce4997f1/internal/identity/oidc/userinfo.go . We use a custom http round tripper that modifies the response to convert the string into a bool so the response follows the spec.

PR #254, opened 2020-06-24
Support signed jwt userinfo response

holowinski:
Hi.

We're using one of the servers which returns `userinfo` response as a signed JWT.
This is, of course, fine according to the spec https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse

This PR adds support for signed JWT `userinfo` responses.

I was wondering how to approach it, and wanted to add it in a way that still works for everyone using this library for `userinfo` fetching without breaking anything.

Added a check based on the `Content-Type` header `application/jwt` which seams reasonable.

Additionally, I've added tests checking that the UserInfo is generated properly from regular JSON and JWT.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -230,6 +230,14 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 		return nil, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 
+	if strings.EqualFold(resp.Header.Get("Content-Type"), "application/jwt") {
+		payload, err := p.remoteKeySet.VerifySignature(ctx, string(body))
+		if err != nil {
+			return nil, fmt.Errorf("oidc: invalid userinfo jwt signature %v", err)
+		}
+		body = payload
+	}
+
 	var userInfo UserInfo
 	if err := json.Unmarshal(body, &userInfo); err != nil {
 		return nil, fmt.Errorf("oidc: failed to decode userinfo: %v", err)
--- a/oidc_test.go
+++ b/oidc_test.go
@@ -9,6 +9,8 @@ import (
 	"reflect"
 	"strings"
 	"testing"
+
+	"golang.org/x/oauth2"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


oidc.go
```diff
@@ -230,6 +231,22 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 		return nil, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 
+	if strings.EqualFold(resp.Header.Get("Content-Type"), "application/jwt") {
+		rawUserInfo := string(body)
+		payload, err := parseJWT(rawUserInfo)
```
ericchiang: parseJWT is used as part of ID token validation so we can verify the claims before having to verify the signature (which is potentially more expansive). Since you're not doing that here, this if block can just be

```
if strings.EqualFold(resp.Header.Get("Content-Type"), "application/jwt") {
		payload, err := p.remoteKeySet.VerifySignature(ctx, string(body))
		if err != nil {
			return nil, fmt.Errorf("oidc: invalid userinfo jwt signature %v", err)
		}
		body = payload
}
```

oidc_test.go
```diff
@@ -297,3 +299,148 @@ func TestNewProvider(t *testing.T) {
...
+			ctx := context.Background()
+
+			provider, err := NewProvider(ctx, mockServer.Server.URL)
+			if err != nil {
+				t.Fatalf("Failed to initialize provider for test %v", err)
+			}
+
+			fakeOauthToken := oauth2.Token{}
+			info, err := provider.UserInfo(ctx, oauth2.StaticTokenSource(&fakeOauthToken))
+
```
ericchiang: nit: remove newline here

oidc_test.go
```diff
@@ -297,3 +299,148 @@ func TestNewProvider(t *testing.T) {
 		})
 	}
 }
+
+type userInfoTestServer struct {
```
ericchiang: Instead of having userInfoTestServer and mockServer as separate structs, just have one "testServer" struct:

```go
type testServer struct {
	wellKnown   string
	contentType string
	jwks        string
	userInfo    string
}

func (t *testServer) run(t *testing.T) string {
    mux := http.NewServeMux()
    // register mux handlers
    srv := httptest.NewServer(mux)
    t.Cleanup(srv.Close)
    return srv.URL
}
```

Your test is then

```go
t.Run(test.name, func(t *testing.T) {
    serverURL := test.server.run(t)

    // logic
})
```

oidc_test.go
```diff
@@ -297,3 +299,148 @@ func TestNewProvider(t *testing.T) {
...
+
+			if info.Email != test.wantUserInfo.Email {
+				t.Errorf("expected UserInfo to be %v , got %v", test.wantUserInfo, info)
+			}
+		})
+	}
+
+}
+
+type mockServer struct {
```
ericchiang: Commented above but I think this struct is redundant.

oidc.go
holowinski: yep, correct.

I've played with this for a bit and assumed that this might be useful for scenarios with invalid JWTs - to fail right away without doing the verification.

But I guess this is too much optimization.
I'll remove it.

oidc_test.go
holowinski: In this case, it's going to start the HTTP server for every test case, which I considered costly.
I wanted to separate the startup from the single test run.

If you don't care about the start cost I can simplify it similar to what you suggested.
Should I ?

ericchiang: I'm fine starting it for every test case. If it's an actual performance issue we can optimize then.

holowinski: Changed it, additionally moved some of the shared mock responses into the `run(t *testing.T)`.

ericchiang (approved):
lgtm, please squash your commits and I'll merge :)

holowinski:
Squashed into one commit.

ericchiang (approved):
lgtm, thanks!

PR #256, opened 2020-06-26
Userinfo - handling charset in Content-Type header

holowinski:
Hi.

This PR fixes a case when the userinfo endpoint responds (properly) with `charset` encoding in the `Content-Type` header.
This case wasn't properly handled before and it came up during tests with one of the authorization servers we're using.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -230,7 +230,9 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 		return nil, fmt.Errorf("%s: %s", resp.Status, body)
 	}
 
-	if strings.EqualFold(resp.Header.Get("Content-Type"), "application/jwt") {
+	ct := resp.Header.Get("Content-Type")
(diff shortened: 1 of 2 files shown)


oidc.go
```diff
@@ -246,6 +246,15 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 	return &userInfo, nil
 }
 
+func mediaType(resp *http.Response) string {
```
ericchiang: You want mime.ParseMediaType

https://play.golang.org/p/7hDblPWcorU

```go
func mediaType(resp *http.Response) string {
    ct := resp.Header.Get("Content-Type")
    if t, _, err := mime.ParseMediaType(ct); err == nil {
        return t
    }
    return ""
}
```

ericchiang (commented):
Sorry, missed this one

oidc.go
holowinski: Thanks. Updated.

ericchiang:
Merged as #261

PR #257, opened 2020-07-15
support AWS Cognito, which is not completely OIDC compliant 

DallanQ:
I've been using go-oidc and needed to support AWS Cognito, which varies from the OIDC spec in a couple of places. This PR makes go-oidc handle the inconsistencies. I'm happy to make changes or add tests. Just wanted to submit it for consideration.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -192,6 +193,16 @@ type UserInfo struct {
 	claims []byte
 }
 
+type userInfoRaw struct {
+	Subject string `json:"sub"`
+	Profile string `json:"profile"`
+	Email   string `json:"email"`
+	// Handle providers that return email_verified as a string
+	// https://forums.aws.amazon.com/thread.jspa?messageID=949441&#949441 and
+	// https://discuss.elastic.co/t/openid-error-after-authenticating-against-aws-cognito/206018/11
+	EmailVerified stringAsBool `json:"email_verified"`
+}
+
 // Claims unmarshals the raw JSON object claims into the provided object.
 func (u *UserInfo) Claims(v interface{}) error {
 	if u.claims == nil {
@@ -240,12 +251,17 @@ func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource)
 		body = payload
 	}
 
-	var userInfo UserInfo
+	var userInfo userInfoRaw
 	if err := json.Unmarshal(body, &userInfo); err != nil {
 		return nil, fmt.Errorf("oidc: failed to decode userinfo: %v", err)
 	}
-	userInfo.claims = body
-	return &userInfo, nil
+	return &UserInfo{
+		Subject:       userInfo.Subject,
+		Profile:       userInfo.Profile,
+		Email:         userInfo.Email,
+		EmailVerified: bool(userInfo.EmailVerified),
+		claims:        body,
+	}, nil
 }
 
 // IDToken is an OpenID Connect extension that provides a predictable representation
@@ -365,6 +381,28 @@ type idToken struct {
 type claimSource struct {
 	Endpoint    string `json:"endpoint"`
 	AccessToken string `json:"access_token"`
+}
+
+type stringAsBool bool
+
+func (sb *stringAsBool) UnmarshalJSON(b []byte) error {
+	var result bool
+	err := json.Unmarshal(b, &result)
+	if err == nil {
+		*sb = stringAsBool(result)
+		return nil
+	}
+	var s string
+	err = json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	result, err = strconv.ParseBool(s)
+	if err != nil {
+		return err
+	}
+	*sb = stringAsBool(result)
+	return nil
 }
 
 type audience []string
--- a/oidc_test.go
+++ b/oidc_test.go
@@ -318,7 +318,7 @@ func (ts *testServer) run(t *testing.T) string {
 				"use": "sig",
 				"kid": "test",
 				"alg": "RS256",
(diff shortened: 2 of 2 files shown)


ericchiang:
Can this be conditioned on a specific issuer URL? E.g. so it only applies to AWS cognito?

oidc.go
```diff
@@ -192,6 +193,14 @@ type UserInfo struct {
 	claims []byte
 }
 
+// UserInfoRaw allows us to decode both compliant and non-compliant Cognito userInfo
+type UserInfoRaw struct {
```
ericchiang: Would prefer to have this unexported so

```
type userInfoRaw struct
```

oidc.go
```diff
@@ -365,6 +380,29 @@ type claimSource struct {
 	AccessToken string `json:"access_token"`
 }
 
+// Cognito reports email_verified as a string in userInfo, so try decoding as a bool or a string
+type StringAsBool bool
```
ericchiang: Same here, this should be

```
type stringBool bool
```

oidc.go
DallanQ: good point. I made it unexported.

DallanQ:
I conditioned the audience check on matching the cognito issuer in Verify. I didn't condition the UserInfo function to unmarshal into either UserInfo or userInfoRaw depending upon the issuer because it would require parsing the token to extract the issuer. But I'm happy to do that if you like. Or I could first unmarshal into UserInfo and if an error is returned, then parse the token, check the issuer, and if the issuer matches cognito then unmarshal into userInfoRaw (which I'd probably want to rename to userInfoCognito then).

oidc.go
DallanQ: I made this change as well.

DallanQ:
i'm happy to make any other improvements.

ericchiang (commented):
Need to add test cases, and please squash your commits:

https://github.com/coreos/go-oidc/blob/bc56623f65f57e200a12533bb7ddad3be8a3dfd9/verify_test.go#L29
https://github.com/coreos/go-oidc/blob/bc56623f65f57e200a12533bb7ddad3be8a3dfd9/oidc_test.go#L358

Otherwise lgtm

ericchiang:
Also would you mind including a link to the AWS Cognito OpenID Connect docs as a comment somewhere in the code? After a bit of searching I can't find anything.

DallanQ:
I added tests, links to documentation, and squashed the commits. There's not a lot of documentation available though. I found these issues through trial and error.

I want to make sure I'm thinking correctly. Auth0 sets aud on access tokens (as well as id tokens) and I've been passing Auth0 access tokens into Verify and verifying the audience. Cognito sets client_id on access tokens but not aud, which is why I added the client_id check for Cognito. (The client_id on a Cognito access token has the same value as aud.) But maybe instead of trying to verify the client ID for an access token I should set the SkipClientIDCheck flag in my code to skip client ID verification when I pass an access token to Verify, in which case I should remove the special-casing for Cognito client_id in Verify? I like the idea of having Verify verify the audience on access tokens, but is that incorrect? I'm not an expert in how to use OIDC.

Also, I couldn't find the public/private keypair you used so I generated a new one and used that for the tests.

ericchiang:
Yes you can do the following without modifying go-oidc, you can even make this logic conditional on AWS Cognito:

```go
c := &oidc. Config{
    SkipClientIDCheck: true,
}
v := provider.Verifier(c)
idToken, err := v.Verify(ctx, rawIDToken)
if err != nil {
    // ...
}
var claims struct {
    ClientID string `json:"client_id"`
}
if err := idToken.Claims(&claims); err != nil {
    // ...
}
if claims.ClientID != clientID {
    // return error
}
```

Would it be better to change this PR to just do the user info fixes and document oidc.Config.ClientID better? Specifically that it verifies the "aud" claim and if you want to skip that validation just provide the oidc.Config.SkipClientIDCheck option? As a library maintainer I'd obvious prefer that since I don't have to deal with providers that go off spec :)

DallanQ:
In hindsight I agree that's the best approach. I reverted the changes to Verify, added a comment, and just left the changes to UserInfo.

verify.go
```diff
@@ -205,6 +205,10 @@ func parseClaim(raw []byte, name string, v interface{}) error {
 //
 //    token, err := verifier.Verify(ctx, rawIDToken)
 //
+// Note that AWS Cognito access tokens include client_id instead of aud, so to verify a
```
ericchiang: Mind dropping this? We don't call out other providers like this. I can follow up with a doc change for the Config struct to specify that it's checking the "aud" claim for the client ID.

oidc.go
```diff
@@ -192,6 +193,18 @@ type UserInfo struct {
 	claims []byte
 }
 
+// userInfoRaw allows us to decode both compliant and non-compliant Cognito userInfo
+// Cognito returns email_verified as a string instead of as a bool.
+// This is not documented by AWS as far as I can tell, but see for example
```
ericchiang: nit: Comments generally aren't in the first person.

Maybe just a comment above EmailVerified

```go
type userInfoRaw struct {
	Subject       string       `json:"sub"`
	Profile       string       `json:"profile"`
	Email         string       `json:"email"`
	// Handle providers that return email_verified as a string.
	// https://forums.aws.amazon.com/thread.jspa?messageID=949441&#949441
	// https://discuss.elastic.co/t/openid-error-after-authenticating-against-aws-cognito/206018/11
	EmailVerified stringAsBool `json:"email_verified"`
}
```

oidc.go
```diff
@@ -367,6 +385,29 @@ type claimSource struct {
 	AccessToken string `json:"access_token"`
 }
 
+// Cognito reports email_verified as a string in userInfo, so try decoding as a bool or a string
```
ericchiang: nit: drop the comment here.

ericchiang (approved):
Few comments, but lgtm otherwise

oidc.go
DallanQ: I deleted the comments above and added the comments above EmailVerified.

oidc.go
DallanQ: done

DallanQ:
Is there anything else you need for this to get merged?

verify.go
ericchiang: Is there a reason this wasn't dropped?

DallanQ: oversight. it's been removed now.

DallanQ:
I believe I have made all of the changes requested.

DallanQ:
Is everything ok with this PR?

DallanQ:
just following up on this PR. Is it ok to merge?

DallanQ:
Thank you!

PR #259, opened 2020-07-17
*: remove Cache-Control checks from key rotation strategy

ericchiang:
This removes the dependency on github.com/pquerna/cachecontrol and
brings the package more inline with the strategy from the OpenID Spec
for rotation.

https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys

Updates https://github.com/coreos/go-oidc/issues/232


115100:
@ericchiang 

Is it possible for you cherry-pick this onto v2 and tag a new release? We're hitting issues with JWK rotation due to the previous behaviour.

I've tried it locally and besides the `go.mod`/`go.sum`, it's a clean cherry-pick. The tests all pass locally too. I can raise a PR for this if you'd like.

PR #274, opened 2020-08-28
Don't store provider context in KeySet

peragwin:
Verify is passed its own context; use this one instead. This avoids any
issue with the providers context going stale.

Since it was effectively only used to store a value for oauth2.HTTPClient,
we can extract this in oidc.NewProvider and pass it explicitly to the
KeySet.

This could potentially cause a subtle bug if users depend on this
specifically being accessed from the context, and are assigning new
http.Clients into the same context instead of modifying the client
instance that already exists, since now only the client stored in the
context at the time of NewProvider will be used.

On the other hand, making this change avoids some undefined behavior,
and storing contexts in structs is arguably an anti-pattern.

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/pquerna/cachecontrol"
+	"golang.org/x/oauth2"
 	jose "gopkg.in/square/go-jose.v2"
 )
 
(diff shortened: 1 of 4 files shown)


peragwin:
Meh.. just saw the other request for the same issue. I don't think this resolves anything in that discussion so closing this..

PR #278, opened 2020-10-16
Special-case Azure

jbchen120:
Azure doesn't follow OIDC specs, so we need to special-case it in the OIDC provider/verifier code.


jbchen120:
Sorry, you can ignore this.

PR #282, opened 2020-12-03
*: remove Cache-Control checks from key rotation strategy

### LINKED ISSUE #272: Remove Cache-Control checks from key rotation strategy
I'm not sure if anyone saw the notification for [this](https://github.com/coreos/go-oidc/pull/259#issuecomment-666371768) so I'm raising an issue separately.

The Cache-Control checks were removed for v3 but the patch applies cleanly onto v2 (besides the `go.mod`/`go.sum` changes); we're hitting an issue in v2 where the returned payload comes with a Cache-Control header and would like this released in v2.

I'll be happy to cherry-pick and raise this PR if everyone's busy.

115100:
This is a cherry-pick of https://github.com/coreos/go-oidc/pull/259 (minus `go.{mod,sum}` changes) onto v2.

We need this as the bug fix is only in the non-tagged v3 branch.

Closes https://github.com/coreos/go-oidc/issues/272.

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -9,19 +9,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pquerna/cachecontrol"
 	jose "gopkg.in/square/go-jose.v2"
 )
-
-// keysExpiryDelta is the allowed clock skew between a client and the OpenID Connect
-// server.
-//
-// When keys expire, they are valid for this amount of time after.
-//
-// If the keys have not expired, and an ID Token claims it was signed by a key not in
-// the cache, if and only if the keys expire in this amount of time, the keys will be
-// updated.
-const keysExpiryDelta = 30 * time.Second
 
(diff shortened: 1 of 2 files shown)


mikedanese:
The reason this was held back is because it does change behavior for existing APIs. What is preventing you from upgrading to v3?

115100:
> The reason this was held back is because it does change behavior for existing APIs.

Agreed, but the current behaviour in v2 essentially means we have to restart whenever our OIDC provider rotates their keys.

>What is preventing you from upgrading to v3?

This is probably preferable but v3 is untagged right now. Can a release be tagged then?

ericchiang:
I sent a PR to update the README (#285), but I think we're basically good to tag v3 otherwise.

ericchiang:
v3's tagged https://pkg.go.dev/github.com/coreos/go-oidc/v3@v3.0.0/oidc

115100:
Amazing, thanks!

PR #284, opened 2020-12-11
Add function to retrive JWKS URL

lucacome:
Added a function to expose JWKS URL.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -69,6 +69,7 @@ type Provider struct {
 	issuer      string
 	authURL     string
 	tokenURL    string
+	jwksURL     string
 	userInfoURL string
 	algorithms  []string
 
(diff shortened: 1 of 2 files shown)


mikedanese:
Code looks good but this would do the trick without expanding the public API of oidc.Provider:

```go
func getJWKSURL(provider *oidc.Provider) (string, error) {
  var claims struct {
    JWKSURL string `json:"jwks_uri"`
  }

  if err := provider.Claims(&claims); err != nil {
    return "", err
  }
  return claims.JWKSURL, nil
}
```

Can you help me understand the motivation for adding a specific method for the `jwks_uri` claim?

ericchiang:
Closing for now. Please let us know if it'd be better to document this.

PR #286, opened 2021-01-07
Various cherry picks from v2 onto v3

ericchiang:
Includes #266, #261, #255, #254

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


oidc/oidc.go
```diff
@@ -378,6 +394,28 @@ type claimSource struct {
...
+	if err != nil {
+		return err
+	}
+	result, err = strconv.ParseBool(s)
+	if err != nil {
+		return err
+	}
+	*sb = stringAsBool(result)
+	return nil
+}
```
mikedanese: This might be simpler:

```go
func (sb *stringAsBool) UnmarshalJSON(b []byte) error {
	switch string(b) {
	case "true", `"true"`:
		*sb = stringAsBool(true)
	case "false", `"false"`:
		*sb = stringAsBool(false)
	default:
		return errors.New("wat")
	}
	return nil
}
```

mikedanese (approved):
Spot checked for anything that looked breaking. LGTM.

oidc/oidc.go
ericchiang: Good idea. Added as a separate commit to preserve commit history.

Issue #288, opened 2021-02-01
Support: Example docs, next steps...


TJM:
I am trying to use this go-oidc v3 in gin-gonic. I started to create "middleware" based on the [examples](https://github.com/coreos/go-oidc/tree/v3/example) included here. This was what I came up with initially: https://github.com/TJM/gin-gonic-oidcauth/tree/basic_insecure.

As the branch name states, that is a basic_insecure example, but it is really not useful for authentication, so...

I then did a bit of research and determined that "nonce" and "state" should definitely not be static strings. It seemed like "state" should be something specific to the client (user's browser), and should thus be stored in the session. The nonce *could* be stored in the session as well, but I found https://pkg.go.dev/github.com/letsencrypt/boulder/nonce which seems to implement "nonce as a service" :p ... and it seems to be working as desired. I am actually just using a nonce for both the nonce and the state. I am not really sure whether thats good or bad :) ... Anyhow, now I am here: https://github.com/TJM/gin-gonic-oidcauth/tree/main ... which I think is a "better" (more secure at least) example, but it is still a contrived example, that dumps a bunch of information into the session, and is able to change from "Hello, world" to "Hello, NAME" ...

Now, I feel like the initial authentication is handled (for the callback), but I am not sure how to handle subsequent requests from the user.

1. What "token" should I store for the user (in the session). The `idToken.AccessToken()` is what I am currently storing, but I am not sure how to validate it. (I guess I could just "assume" that the encrypted/authenticated session has not been tampered with, but that seems bad?)
2. Should I do any token validation on subsequent requests, or just store the "iat" and "exp" (currently storing all claims), and validate that they are still valid.
3. Are there any examples about "refresh" tokens? (renewing the token from the app while the user is still active)

Thanks in advance,
Tommy

ericchiang:
Hey Tommy,

These seem like general questions about OAuth 2.0 security and session management, not necessarily a bug in this library.

Unfortunately, the maintainers of this package generally don't have bandwidth to answer support questions like this. Below are some links that might help getting started:

* https://auth0.com/docs/protocols/state-parameters
* https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html
* https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes

Hope this helps,
Eric

TJM:
[...]

PR #289, opened 2021-02-04
example: use real cookies for state and nonce tokens

ericchiang:
This also consolidates the nonce  example into the "idtoken" app.

@TJM is this what you had in mind with https://github.com/coreos/go-oidc/issues/288?


TJM:
I like it! Thank you!

For what its worth, I used the "NonceService" from https://github.com/letsencrypt/boulder/tree/main/nonce ... but implementing nonce might be out of scope for the example. It is a much, much better example now. Thanks!

https://github.com/TJM/gin-gonic-oidcauth/blob/main/oidc.go#L270

~tommy

PR #291, opened 2021-03-12
Update jwks doc about cache keys

spk:
Hi from https://github.com/coreos/go-oidc/commit/8858d8afe4f964bcbc26a582d84dff7548d177c4 the cache-control is not used for checks from key rotation strategy, let me know if you have better suggestion for the text, thanks


ericchiang (approved):
thanks!

PR #292, opened 2021-03-12
verify.go: return meaningful error types

### LINKED ISSUE #283: Allow verification errors to be differentiated
Right now the package uses fmt.Errorf() for all errors. For example:
```
return nil, fmt.Errorf("oidc: token is expired (Token Expiry: %v)", t.Expiry)
```
This prevents callers from being able to identify individual errors. For example, a middleware wanting to return "Bad credentials." for most generic failures, but "Token expired." specifically for that one error. Right now I am forced to do a strings.Contains to identify a token expiration error.

My suggestion is to define sentinel errors (e.g ErrTokenExpired) for cases where the additional context is irrelevant*,  and to define custom error types (e.g. TokenExpiredError) where the context is relevant.

* I'd argue that including t.Expiry in this error isn't super helpful cause it's my token and I already know the expiration, but it's definitely subjective.

intinig:
In order to avoid clients the need to parse strings to understand what
happened to a client, and make token verification more robust,  I added
several error types that are returned instead of a generic fmt.Errorf
message.

Had to change a little how the test works in order to test for an
expected return type, but everything seems to work for the best

Fixes #283

### DIFF
--- a/oidc/error_types.go
+++ b/oidc/error_types.go
@@ -87,7 +87,7 @@ type ExpiredTokenError struct {
 
 // Error interface
 func (e *ExpiredTokenError) Error() string {
-	return fmt.Sprintf("oidc: token ss expired (Token Expiry: %v)", e.Expiry)
+	return fmt.Sprintf("oidc: token is expired (Token Expiry: %v)", e.Expiry)
 }
 
 // TokenNotYetValidError is returned when a token supplies NotBefore but is dated
--- a/oidc/error_types.go
+++ b/oidc/error_types.go
@@ -1,0 +1,150 @@
+package oidc
+
+import (
+	"fmt"
+	"time"
+)
+
+// MalformedJWTError is returned when the JWT can't be parsed
+type MalformedJwtError struct {
+	ParseError error
+}
+
+// Error interface
+func (e *MalformedJwtError) Error() string {
+	return fmt.Sprintf("oidc: malformed jwt: %v", e.ParseError)
+}
+
+// MalformedPayloadError is returned when it's impossible to unmarshal the
+// JWT payload
+type MalformedPayloadError struct {
+	UnmarshalError error
+}
+
+// Error interface
+func (e *MalformedPayloadError) Error() string {
+	return fmt.Sprintf("oidc: failed to unmarshal claims: %v", e.UnmarshalError)
+}
+
+// InvalidClaimNameError is returned when a claim name returns no claim source
+type InvalidClaimNameError struct {
+	Name string
+}
+
+// Error interface
+func (e *InvalidClaimNameError) Error() string {
+	return fmt.Sprintf("oidc: failed to obtain source from claim '%s'", e.Name)
+}
+
+// InvalidClaimSourceError is returned when a claim name references a
+// non existing source
+type InvalidClaimSourceError struct {
+	Name   string
+	Source string
+}
+
+// Error interface
+func (e *InvalidClaimSourceError) Error() string {
+	return fmt.Sprintf("oidc: source '%s', referenced by '%s', does not exist", e.Source, e.Name)
+}
+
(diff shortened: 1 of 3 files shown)


ericchiang:
Thanks for the PR!

I think this exposes way too much internal logic that might change or be reordered. We don't want users to depend on our JWT parsing details, for example. What's the actual problem you're trying to solve? In #283 I suggested a specific error so API users could tell if a token was expired.

intinig:
Thx for the reply I just saw this, so forgive my late one. 

Why do you say it exposes too much internal logic? It's just encapsulating all the potential errors during the Verify process into their own types that can then be used by users to understand what's going on. 

With the exception of a couple of error types (thinking about the malformed payload and malformed token ones here) all of them are errors that are interesting for the users as they expose misconfigurations on their part or forgeries or other things that need action on their part.

This approach saves users from needing to parse the error message returned by the generic error.

ericchiang:
> Why do you say it exposes too much internal logic? It's just encapsulating all the potential errors during the Verify process into their own types that can then be used by users to understand what's going on.

What if we reorder some of this logic? That means the same token could return different errors which is now an API change.

> With the exception of a couple of error types (thinking about the malformed payload and malformed token ones here) all of them are errors that are interesting for the users as they expose misconfigurations on their part or forgeries or other things that need action on their part.

Bare in mind that our "quick checks" execute before signature validation for performance reasons. But if someone is now depending on the fact the that signature validation error is meaningful, we'd want to perform that first. Just another example of exposing internal details.

> This approach saves users from needing to parse the error message returned by the generic error.

Sure, but what scenarios would someone actually use this for? Types errors are predominantly used to have programs behave differently for different failure cases. If you're just going to be logging this for someone to view anyway, then what benefit does this PR gain?

I'm happy to add typed errors, but I'd like there to be a use case so we're not just adding new API surface that doesn't get used, or used in ways we're not comfortable supporting.

What's your use case here? How are you using go-oidc?

intinig:
In our specific use case we have an API that supports multiple signers, multiple client_id per signers and tokens that can expire and need refreshing. 

Knowing what type of error is happening during verification allows us to choose the right action for a specific token, like returning an error, trying the next signer, trying another client id and so on.

The order of execution is not really a problem for us as if a token has multiple errors we will still want to handle all of them, independently of the order in which we're triggering them.

ericchiang:
I think this is a good example: should we return a client_id mismatch or a signature mismatch for an individual token? Does it matter if we returned one kind of error today and another tomorrow? Do we have to do the signature validation before checking the client_id so an attacker can't fake a client_id mismatch?

This is the big reason we added the SkipClientIDCheck, SkipExpiryCheck, SkipIssuerCheck options. If you've got interesting logic here, it might be easier to use those options and perform your own validation.

https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Config

This API is designed intentionally to not let you make decisions before validating the signature. It's super dangerous do that, and has resulted in vulnerabilities in other libraries that process JWTs.

intinig:
It technically allows you to do that processing anyway, cause it returns error messages that explain what's going on.

I just made that a little more explicit :)

I see your point of view, I don't fully agree with it because I'd rather have the validation process tell me what's wrong than do it multiple time skipping a check every time or parsing a text message.

Feel free to reject the PR or adopt part of it, it's a little of a bummer for us cause we'd rather use a mainline library than a fork, but that's part of the beauty of open source :)

Thx again for your time!

ericchiang:
Thanks for your replies during this process! Apologies if anything came off as disregarding the use case, but JWTs are awful.

I'm going to close this given concerns about pre-validation decisions in https://github.com/coreos/go-oidc/pull/292#issuecomment-802232181

Error strings are definitely outside the scope of any API compatibility. We may change or reorder them in the future outside of a major version update.

PR #295, opened 2021-04-06
no-op: remove dead code and minor IDE warnings

enj:
Signed-off-by: Monis Khan <mok@vmware.com>

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -17,7 +17,6 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
-	jose "gopkg.in/square/go-jose.v2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


oidc/oidc.go
```diff
@@ -131,7 +125,7 @@ func NewProvider(ctx context.Context, issuer string) (*Provider, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
ericchiang: nit: Never seen this one? Also seems like it's not what the standard library does?

https://github.com/golang/go/search?q=%22defer+resp.Body.Close%22

ericchiang (approved):
Thanks! Lgtm expect the defer changes

oidc/oidc.go
enj: Goland complains about this being a potentially unhandled error.  I reverted those changes.

ericchiang (approved):

PR #303, opened 2021-05-27
Added all standard claims to UserInfo parsing

dagdagdag83:
Per https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims there were several possible standard claims that can be returned by userinfo endpoint, which this library did not map and did not return.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -189,22 +189,54 @@ func (p *Provider) Endpoint() oauth2.Endpoint {
 
 // UserInfo represents the OpenID Connect userinfo claims.
 type UserInfo struct {
-	Subject       string `json:"sub"`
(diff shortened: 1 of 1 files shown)


ericchiang:
Thanks for the PR!

I think we're going to skip this and stick to only required claims. Most providers don't return this information, and probably shouldn't for privacy reasons. Also if any provider return non-complaint versions of these claims (e.g. we've seen some return email_verified as a string) then this change will now break code that was working before.

This is easy to do yourself with the UserInfo.Claims() method and get the exact claims you're interested in:

https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#UserInfo.Claims

Going to close this out. But if you feel that a docs improvement would help, please feel free send a PR with an example.

PR #308, opened 2021-06-23
Allowing Extention of Remaing OpenId Configuration fields

shantanubansal:


ericchiang:
Similar to https://github.com/coreos/go-oidc/pull/303#issuecomment-849977121, we're not going to add this. Feel free to use the Claims method on the Provider struct to access these fields.

PR #315, opened 2021-09-15
oidc: add option to override discovered issuer URL

### LINKED ISSUE #212: Document Azure "{tenantid}" discovery quirks
This has become such a common question that I think it's worth documenting or having a special error message.

https://github.com/coreos/go-oidc/issues/121
https://github.com/coreos/go-oidc/issues/159
https://github.com/coreos/go-oidc/issues/204

### LINKED ISSUE #250: Issuer Name Mismatch
Hello,

I understand this request was previously declined already: https://github.com/coreos/go-oidc/issues/240

But sometimes there are technical or domain reasons to allow a different `issuer` than the one returned by the server. The standard is not covering this scenario. If you look at the forks of this library, just pick the recent 20 forks, most of them are only doing the change to allow a different issuer. By not allowing this option the eco-system will diverge, we are basically requested to fork the library and go in a different direction.

I hope you can understand the need we have to do this change. Please consider adding a separate option to `NewProvider` to provide the expected issuer. That should be at least as secure or safe as the current implementation.

Sincerely threez

ericchiang:
Fixes https://github.com/coreos/go-oidc/issues/250
Fixes https://github.com/coreos/go-oidc/issues/212

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -39,6 +39,10 @@ var (
 	errInvalidAtHash = errors.New("access token hash does not match value in ID token")
 )
 
+type contextKey int
+
+var issuerURLKey contextKey
+
 // ClientContext returns a new Context that carries the provided HTTP client.
 //
 // This method sets the same context key used by the golang.org/x/oauth2 package,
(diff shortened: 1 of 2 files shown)


liafizan (commented):
We tried the fix in our environment where our OIDC connect provider has the issuer mismatch problem and this fix worked perfectly! Thank you for working on this

ericchiang:
Glad this works for you! I think that's the feedback I need. Going to go ahead and merge.

PR #316, opened 2021-09-20
Increase nbf-leeway to 5 minutes

cgostuff:
The default clock skew for MS JWT validation is 5 minutes. Increasing the default nbf-leeway here would match that: [https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.12.2/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L149-L153](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.12.2/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L149-L153)

### DIFF
--- a/oidc/verify.go
+++ b/oidc/verify.go
@@ -274,7 +274,9 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		// If nbf claim is provided in token, ensure that it is indeed in the past.
 		if token.NotBefore != nil {
 			nbfTime := time.Time(*token.NotBefore)
-			leeway := 1 * time.Minute
+			// Set to 5 minutes since this is what other OpenID Connect providers do to deal with clock skew.
+			// https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.12.2/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L149-L153
+			leeway := 5 * time.Minute
 
 			if nowTime.Add(leeway).Before(nbfTime) {
 				return nil, fmt.Errorf("oidc: current time %v before the nbf (not before) time: %v", nowTime, nbfTime)


ericchiang:
The config takes a time function

https://pkg.go.dev/github.com/coreos/go-oidc/v3@v3.1.0/oidc#Config

I also believe SkipExpiryCheck will let you do your own logic here.

Clock skews more than a minute seem like it'd be better to fix the clock skew. E.g. things like TLS validation might fail as well. What value were you hoping to set this to? Can you sync your ntpd before starting your Go program instead?

cgostuff:
Thanks for the reply. Passing in a custom time function won't affect the nbf-constraint from what I can see.

Identityserver4 uses a default clock skew value of 5 minutes, so I figured it made sense that it's possible to configure this Go-component to a similar level.

I suppose I could use SkipExpiryCheck and do a completely custom expiry check after running the standard Verify function.

ericchiang:
Do you have a link for the identityserver4 skew? Does increasing the skew to 5 minutes work for you, or would you need to set it hire? I'd be open to 5min if there are other components doing that.

cgostuff:
This link refers to the default value: [https://github.com/IdentityServer/IdentityServer4.AccessTokenValidation/issues/90](https://github.com/IdentityServer/IdentityServer4.AccessTokenValidation/issues/90).
Setting the nbf leeway to 5 min does work for me, I wouldn't need it any higher than that.

ericchiang:
If you'd like to update this PR to make the default 5 minutes, I think that'd be fine. Please include the following link in the comment:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.12.2/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L149-L153

I don't think we want to make this configurable. If you need more than 5 minutes, you probably want to fix the clock skew.

cgostuff:
I updated the OP with the link you referred to (hopefully this is what you meant), as well as the title to more accurately reflect the new change.

oidc/verify.go
```diff
@@ -274,7 +274,7 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		// If nbf claim is provided in token, ensure that it is indeed in the past.
 		if token.NotBefore != nil {
 			nbfTime := time.Time(*token.NotBefore)
-			leeway := 1 * time.Minute
+			leeway := 5 * time.Minute
```
ericchiang, line 279: Can you please include a comment for how this was picked?
```
// Set to 5 minutes since this is what other OpenID Connect providers do to deal
// with clock skew.
//
// https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.12.2/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L149-L153
leeway := 5 * time.Minute
```

ericchiang (approved):
lgtm with one nit

Can you squash your commits? (there are 4 now, there should be one)

https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

oidc/verify.go
cgostuff: Added the comment, but messed the squash up.. I suppose I used a number too high in HEAD~X and now I'm not quite sure how to fix it. Any tips?

ericchiang: Maybe try "reset --soft" and reauthor the commit? Something like:

```
$ git remote -v
origin  https://github.com/cgostuff/go-oidc (fetch)
origin  https://github.com/cgostuff/go-oidc (push)
upstream        https://github.com/coreos/go-oidc (fetch)
upstream        https://github.com/coreos/go-oidc (push)
$ git checkout origin/v3
HEAD is now at 5d613d0 Merge branch 'v3' of https://github.com/cgostuff/go-oidc into v3
$ git reset --soft upstream/v3
$ git status
HEAD detached from origin/v3
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   oidc/verify.go

$ git commit -m "Make leeway for the nbf-constraint configurable"
[detached HEAD dc8d817] Make leeway for the nbf-constraint configurable
 1 file changed, 3 insertions(+), 1 deletion(-)
$ git log -n 2
commit dc8d8177cfe75db23906befb62d442a25dce5eaf (HEAD)
Author: Eric Chiang <ericchiang@google.com>
Date:   Fri Oct 1 18:05:25 2021 +0000

    Make leeway for the nbf-constraint configurable

commit d42db69c79f2fa664fd4156e939bf27bba0d2f68 (tag: v3.1.0, upstream/v3.0.0, upstream/v3, origin/v3.0.0)
Merge: 15b94d9 916b64f
Author: Eric Chiang <ericchiang@google.com>
Date:   Thu Sep 16 17:06:04 2021 -0700

    Merge pull request #315 from ericchiang/issuer

    oidc: add option to override discovered issuer URL
 
 ```

cgostuff: Ty for the reply. At "git reset --soft upstream/v3" I get:
fatal: ambiguous argument 'upstream/v3': unknown revision or path not in the working tree.

ericchiang:
Upstream is the remote. You'll need to add it:

Something like

git add remote https://github.com/coreos/go-oidc

cgostuff:
Somehow it has the same exact effect.. I reset to the last commit from the original repository, redo the commit, then it just re-pushes the same changes together with a merge instead of overwriting the history/squashing.

ericchiang:
Sorry this is taking a while. Would you like me to send the change?

cgostuff:
If you can send it, that would be great. I apologize for the hassle, I have relied on AD to squash for me in the past. Edit: the branch should be good to go now.

cgostuff:
Looks like I finally got it! Thanks for guiding me through this @ericchiang.

Issue #317, opened 2021-09-21
Expose whether UserInfo endpoint is supported by provider


adombeck:
I want to use the `UserInfo` endpoint if the provider supports it. If it does not support it, I want to use the `IdToken` instead. Currently, the package provides no way to check if the endpoint provider supports the `UserInfo` endpoint, except calling `Provider.UserInfo()` and then checking if the error message is  `"oidc: user info endpoint is not supported by this provider"`, which would fail if the error message changes in the future.

I see two solutions to this:
1. Export the error that is returned by `Provider.UserInfo()` when `userInfoURL` is empty
2. Export the `userInfoURL` field of the `Provider` struct

ericchiang:
You can use the Claims method to do this:

https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims

Something like

```
p, err := oidc.NewProvider(ctx)
if err != nil {
  // ...
}
var claims struct {
  UserInfoURL string `json:"userinfo_endpoint"`
}
if err := p.Claims(&claims); err != nil {
  // ...
}
```

Maybe we should document this better?

adombeck:
That works for me, thanks!

PR #320, opened 2021-09-23
Session EndPoint Fetch (Usually used alot with Public clouds)

shantanubansal:


ericchiang:
See https://github.com/coreos/go-oidc/issues/317#issuecomment-924083381

You can do this with the Claims method :)

PR #323, opened 2021-11-26
Add device authorization details to Provider info

kellyma2:
If device authorization is supported by the underlying provider we
should be able to retrieve the details for this as well as the auth
endpoint for the regular authorization flow.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -98,11 +98,12 @@ func doRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
 
 // Provider represents an OpenID Connect server's configuration.
 type Provider struct {
-	issuer      string
-	authURL     string
-	tokenURL    string
-	userInfoURL string
-	algorithms  []string
+	issuer        string
+	authURL       string
+	deviceAuthURL string
+	tokenURL      string
+	userInfoURL   string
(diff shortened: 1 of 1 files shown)


ericchiang:
I'd prefer values like this to be accessed through the provider's Claims method. Does that work for your use case?

kellyma2:
Likely it can work, but is there any particular reason why?

ericchiang:
There are a ton of fields returned by the provider object. Adding new API for each doesn't scale, particularly for non-mandatory ones

kellyma2:
UserInfo is optional, but it's already in there as well?

ericchiang:
Needed for other APIs :)

https://github.com/coreos/go-oidc/blob/v3.1.0/oidc/oidc.go#L246

Going to close since it sounds like you can get this information already with the Claims() method. Apologies if this seems like a hard stance, but I'd prefer to have one API for this rather than many

kellyma2:
Thanks for the info.  

I hope you reconsider - UserInfo() is pretty accessor for what would otherwise be an annoying, but optional, API.  Similarly, returning the bits needed for device auth in a clean package makes things a bit nicer for what I'd expect is a common use case.  NBD either way.

PR #326, opened 2021-12-14
Add configurable clock skew

cgostuff:
This implements clock skew similar to [AzureAD](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6.12.2/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L149-L153%20clockSkew%20:=%20time.Duration(5)%20*%20time.Minute). The existing solution with 5 min nbf leeway and a configurable time function does not let you get a true 5 min clock skew, because the time.now() function used to check expiry is reused to check the nbf-constraint.

This is configurable in .NET [TokenValidationParameters.ClockSkew Property](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.tokens.tokenvalidationparameters.clockskew?view=azure-dotnet), and is used like this in IdentityServer4 [PrivateKeyJwtSecretValidator.cs#L106](https://github.com/IdentityServer/IdentityServer4/blob/3ff3b46698f48f164ab1b54d124125d63439f9d0/src/IdentityServer4/src/Validation/Default/PrivateKeyJwtSecretValidator.cs#L106).

### DIFF
--- a/oidc/verify.go
+++ b/oidc/verify.go
@@ -100,6 +100,9 @@ type Config struct {
 
 	// Time function to check Token expiry. Defaults to time.Now
 	Now func() time.Time
+
+	// Duration for clock skew. Defaults to 5 minutes.
+	ClockSkew time.Duration
 }
 
 // Verifier returns an IDTokenVerifier that uses the provider's key set to verify JWTs.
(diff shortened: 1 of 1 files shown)


jordaniversen:
Any word on this?  I was also looking for such a feature.

ericchiang:
Closing since if you're clock is that skewed, you've got much bigger issues and this will impact all sorts of cryptography throughout your program (e.g. TLS certs). I think my suggestion will be providing your own time.Now function, or fixing your system clock

PR #328, opened 2022-02-02
Add minor error handling

loosebazooka:
Not sure that this is strictly necessary here but helps debug a bit: adds an error if an empty `jwks_uri` is provided. Although maybe that's allowed?


mitar:
It would be great if this fix is merged.

PR #336, opened 2022-05-10
Remove obsolete documentation

### LINKED ISSUE #335: IDTokenVerifier from Provider
`Provider.Verifier`'s documentation currently states:

> The returned IDTokenVerifier is tied to the Provider's context and its behavior is undefined once the Provider's context is canceled. 

Looking at the code I do not see this to be true. There is no context being passed from `Provider` to `IDTokenVerifier`. There is no context even saved into `Provider`. The only context is inside `RemoteKeySet` but even that one explicitly creates a new/cloned context using `context.Background()`.

I think this line should be removed from documentation.

mitar:
Fixes #335.

### DIFF
--- a/oidc/verify.go
+++ b/oidc/verify.go
@@ -103,9 +103,6 @@ type Config struct {
 }
 
 // Verifier returns an IDTokenVerifier that uses the provider's key set to verify JWTs.
-//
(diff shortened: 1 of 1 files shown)


ericchiang:
thanks!

PR #337, opened 2022-05-11
oidc: add StaticKeySet

ericchiang:
This has been extremely useful for testing, but I've been hesitant to
add any API surface that references our underlying JOSE library.

Add a StaticKeySet that only accepts keys of type *rsa.PublicKey and
*ecdsa.PublicKey.

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -2,6 +2,9 @@ package oidc
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


ericchiang:
cc @mitar

mitar:
Awesome! This looks great and I can use it as-is.

PR #340, opened 2022-06-14
Support ed25519 keys

chancez:
Adds support for ed25519 keys

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/rsa"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


chancez:
CC @ericchiang

ericchiang:
Hey chance!

I think this package would still reject Ed25519 signed tokens because we only permit allow listed signature types in the JWT header.

Are you working from a spec somewhere?

chancez:
Hey!

I hadn't considered that it might not be in the spec. I checked, and all I can find is this https://datatracker.ietf.org/doc/html/rfc8037 which defines how it works with JOSE.

At the moment, I'm trying to convert some home-grown OIDC code in a project I'm working on to go-oidc and we were using https://github.com/cristalhq/jwt previously which supports ed25519. As far as I can tell, I think the Ed25519 use-case was for handling JWTs signed by a certificate, such as the K8s service account tokens. 

So far everything works if I include this change, go-jose supports it. Previously we had a `jwt.Verifier` interface everywhere and I swapped it with the `*oidc.IDTokenVerifier`, but it seems I made a mistake assuming I could reuse the ID token verifier for non-ID token based JWTs (it was a bit of a hack, but it was simpler than other methods I started with). 

I can look at alternative routes if you don't think this can be accepted.

chancez:
Okay I've got myself a different solution so that this isn't necessary. 

I had to create a few interfaces to handle using go-jose vs go-oidc based on the use-case, but it's all working now. Feel free to close this PR if you think it's not something that can be accepted.

ericchiang:
Okay sounds good. Feel free to reopen if you need this in the future!

PR #342, opened 2022-06-23
feat: export token expired error

### LINKED ISSUE #327: What is the appropriate way to handle the errors returned from IDTokenVerifier.Verify()?
Well to give some context, I am using go-oidc in a middleware to verify JWT token. *oidc.IDTokenVerifier.Verify() returns IDToken and error. The verify function could possibly return different errors. Depending on the errors I want to return appropriate HTTP status code to the user. But it looks like the pkg does not expose error types. 

In the source code I see the errors are returned like below:
`fmt.Errorf("oidc: malformed jwt: %v", err) or fmt.Errorf("oidc: source does not exist")` etc.

So I am not sure how to compare these errors and return appropriate HTTP status code.
For now I am using `strings.Contain()` to compare, but I don't know if there is any better way to do this.

crenshaw-dev:
Relates to #327 

This allows `errors.Is` checking while preserving the exact same error messages for those who are doing string comparison.

Preserving the exact same strings means some of the error messages (without context) look a little weird. But I think it's a reasonable tradeoff until it's time for another major version bump.

This PR would let me write unit tests in Argo CD without worrying that a change in go-oidc will break my string comparison checks.

### DIFF
--- a/oidc/verify.go
+++ b/oidc/verify.go
@@ -21,6 +21,18 @@ const (
 	issuerGoogleAccountsNoScheme = "accounts.google.com"
 )
 
+// TokenExpiredError indicates that Verify failed because the token was expired. This
+// error does NOT indicate that the token is not also invalid for other reasons. Other
+// checks might have failed if the expiration check had not failed.
+type TokenExpiredError struct {
+	// Expiry is the time when the token expired.
+	Expiry time.Time
+}
+
+func (e *TokenExpiredError) Error() string {
+	return fmt.Sprintf("oidc: token is expired (Token Expiry: %v)", e.Expiry)
+}
+
 // KeySet is a set of publc JSON Web Keys that can be used to validate the signature
 // of JSON web tokens. This is expected to be backed by a remote key set through
 // provider metadata discovery or an in-memory set of keys delivered out-of-band.
@@ -260,7 +272,7 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		nowTime := now()
 
 		if t.Expiry.Before(nowTime) {
-			return nil, fmt.Errorf("oidc: token is expired (Token Expiry: %v)", t.Expiry)
+			return nil, &TokenExpiredError{Expiry: t.Expiry}
 		}
 
 		// If nbf claim is provided in token, ensure that it is indeed in the past.
--- a/oidc/verify_test.go
+++ b/oidc/verify_test.go
@@ -3,6 +3,7 @@ package oidc
 import (
 	"context"
 	"crypto"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -72,8 +73,8 @@ func TestVerify(t *testing.T) {
 			config: Config{
 				SkipClientIDCheck: true,
 			},
-			signKey: newRSAKey(t),
-			wantErr: true,
+			signKey:   newRSAKey(t),
+			wantErrAs: &TokenExpiredError{},
 		},
 		{
 			name:    "unexpired token",
@@ -530,8 +531,9 @@ type verificationTest struct {
 	// testing invalid signatures.
 	verificationKey *signingKey
 
-	config  Config
-	wantErr bool
+	config    Config
+	wantErr   bool
+	wantErrAs error
 }
 
 func (v verificationTest) runGetToken(t *testing.T) (*IDToken, error) {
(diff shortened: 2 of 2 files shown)


ericchiang:
Hey @crenshaw-dev,

I don't want to export every possible error. If you've got a specific use case where a typed error would be helpful (e.g. specific HTTP status codes from the provider), happy to add that. Do you have an example of what you're trying to do?

Also, some of these errors arguably shouldn't be used problematically, since they're hit before we validate the signature check on the ID token. 

Eric

crenshaw-dev:
@ericchiang there are two current use cases in Argo CD:

1) We want to know if it's a "token expired" error so the UI can redirect the user immediately to a new login flow: https://github.com/argoproj/argo-cd/blob/a588e63e3cb9db91fd491c0d7371c87867861aa7/util/session/sessionmanager.go#L493-L504
2) We have a regression test to make sure we support algorithms besides RS256 (added support with 2.2.0): https://github.com/argoproj/argo-cd/blob/a588e63e3cb9db91fd491c0d7371c87867861aa7/util/session/sessionmanager_test.go#L581

It's possible that test shouldn't exist. But I think the expired example is a reasonable use case. Would you be interested in a PR that only exports that error?

ericchiang:
I think a token expired error would be reasonable to add!

That second test probably shouldn't be necessary? Just make sure your RS512 token doesn't return an error.

So sounds like the API we'd want to add it

```
var ErrTokenExpired = errors.New(...)
```

(nit: exported error variables generally start with "Err", and error types end with "Error" https://pkg.go.dev/io#pkg-variables, https://pkg.go.dev/net#AddrError)

Kinda a tangent, but I've also been thinking about having an oidctest package that would let you stand up a httptest provider with some basic configuration. Maybe that would be helpful too.

crenshaw-dev:
> Just make sure your RS512 token doesn't return an error.

Yeeeeah that would be proper. I was lazy and didn't want to have to set up a full mock OIDC provider. Maybe that would be easier than I assumed.

> (nit: exported error variables generally start with "Err", and error types end with "Error" https://pkg.go.dev/io#pkg-variables, https://pkg.go.dev/net#AddrError)

Good catch, will fix!

> Kinda a tangent, but I've also been thinking about having an oidctest package that would let you stand up a httptest provider with some basic configuration. Maybe that would be helpful too.

Yep, that would make some other tests I'm working on a lot easier I think! Maybe I can put some Intuit time towards that eventually.

Updating the PR to be more minimal!

oidc/verify.go
```diff
@@ -13,14 +13,19 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
-	jose "gopkg.in/square/go-jose.v2"
```
ericchiang: nit: please leave this line alone. I think adding this is the default goimports functionality.

oidc/verify.go
```diff
@@ -13,14 +13,19 @@ import (
...
 
 const (
 	issuerGoogleAccounts         = "https://accounts.google.com"
 	issuerGoogleAccountsNoScheme = "accounts.google.com"
 )
 
+// TokenExpiredErr indicates that Verify failed because the token was expired. This
+// error does NOT indicate that the token is not also invalid for other reasons. Other
+// checks might have failed if the expiration check had not failed.
+var TokenExpiredErr = errors.New("oidc: token is expired")
```
ericchiang: nit: This should be ErrTokenExpired

Or maybe you want to include the expiry?

```
type TokenExpiredError struct {
    Expiry time.Time
}
```

ericchiang (commented):
(please squash your commits)

oidc/verify.go
crenshaw-dev: Sure thing. Gonna fight with GoLand real quick.  :-)

oidc/verify.go
crenshaw-dev: I like that

oidc/verify_test.go
```diff
@@ -12,6 +13,9 @@ import (
 	"time"
 )
 
+// anyError is a fake error to match any error returned in a test.
+var anyError = errors.New("")
```
crenshaw-dev: This may or may not be the proper way to do this. lmk

crenshaw-dev:
@ericchiang ready for another look

oidc/verify_test.go
ericchiang: Add a new field and use that instead?

```
type verificationTest struct {
	wantErr bool
	
	wantErrIs error
}
```

oidc/verify.go
```diff
@@ -21,6 +21,25 @@ const (
...
+	Expiry time.Time
+}
+
+func (e *TokenExpiredError) Error() string {
+	return fmt.Sprintf("oidc: token is expired (Token Expiry: %v)", e.Expiry)
+}
+
+// Is returns true if the error was due to an expired token. It does not check that
+// the expiry time is identical.
+func (e *TokenExpiredError) Is(target error) bool {
```
ericchiang: Implementing is shouldn't be necessary. Can you drop this? https://go.dev/play/p/fTGu3-8H819

ericchiang (approved):

oidc/verify.go
crenshaw-dev: Apparently `errors.Is` is doing string comparison. Since the Expiry is included in the time, I have to use a custom `Is` to avoid checking the Expiry.

crenshaw-dev: Demo: https://go.dev/play/p/7rN5_y9ODKD

ericchiang: You use errors.As instead for error types, not errors.Is:

```
var expErr *oidc.TokenExpiryError
if errors.As(err, &expErr) {
    // ...
}
```

Would recommend dropping Is() implementation. That's generally reserved for logic that needs to be very custom.

crenshaw-dev:
@ericchiang updated

ericchiang:
thanks!

PR #349, opened 2022-09-07
Adding a way to get the end session endpoint

zuchi:
To be able to end up the session this pull request aims to create a public function to be able to get which will be the end session endpoint based on the discovery (.well-known) provided by the IDP.

Please verify if it makes sense for you and cases it does, let's get it merged. Thank you so much.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /bin
 /gopath
+.idea/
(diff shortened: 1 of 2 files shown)


ericchiang:
You can already do this through the Claims() method :)

https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims

Since there are a lot of field in the well-known endpoint, I'd prefer users stick to Claims instead of adding new APIs for each

PR #352, opened 2022-09-12
Update Issuer verification to skip last useless slash

Flatch78:
Some OIDC solution provider return the Issuer in openid-configuration with useless slash at the end.
That fix allows to skip the last slash during Issuer compare.


ericchiang:
Dup of 

https://github.com/coreos/go-oidc/issues/203
https://github.com/coreos/go-oidc/issues/238

PR #363, opened 2023-01-13
http_requests: use context in http requests

### LINKED ISSUE #362: Use context in new requests
In some methods the code below do not use the context passed thought the method:
ex:

```
func (p *Provider) UserInfo(ctx context.Context, tokenSource oauth2.TokenSource) (*UserInfo, error) {
	...
	if p.userInfoURL == "" {
		return nil, errors.New("oidc: user info endpoint is not supported by this provider")
	}

	req, err := http.NewRequest("GET", p.userInfoURL, nil)
	...

```

There is something that create some kind of risk in using :
`	req, err := http.NewRequestWithContext(ctx, "GET", p.userInfoURL, nil)`

?

Thanks.

argenta-quanto:
this use the context sent to methods in requests

Fixes #362

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -219,7 +219,7 @@ func (r *RemoteKeySet) keysFromRemote(ctx context.Context) ([]jose.JSONWebKey, e
 }
 
 func (r *RemoteKeySet) updateKeys() ([]jose.JSONWebKey, error) {
-	req, err := http.NewRequest("GET", r.jwksURL, nil)
+	req, err := http.NewRequestWithContext(r.ctx, "GET", r.jwksURL, nil)
 	if err != nil {
 		return nil, fmt.Errorf("oidc: can't create request: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


ericchiang:
Contexts are already associated with requests here: https://github.com/coreos/go-oidc/blob/a8ceb9a2043fca2e43518633920db746808b1138/oidc/oidc.go#L96

Is there a reason to have this multiple places?

argenta-quanto:
> Contexts are already associated with requests here:
> 
> https://github.com/coreos/go-oidc/blob/a8ceb9a2043fca2e43518633920db746808b1138/oidc/oidc.go#L96
> 
> Is there a reason to have this multiple places?

I'm using a custom client to use a custom transport that generate some metrics in every requests my application does, and for some reason the request to /jwks do not have my context. 

But I see the  **return client.Do(req.WithContext(ctx))**  you tell me know.

Thanks for response, and i'm closing the PR and the issue. Maybe some problem in my code.

Thanks.

argenta-quanto:
I suspect that the **cloneContext(ctx)** in **newRemoteKeySet** is using a context.BackGround() with only the client. So its ignoring my original context another values.

PR #364, opened 2023-01-13
oidc: restrict use of context.Background()

### LINKED ISSUE #339: Non-obvious behavior with long lived `oidc.Provider` objects: `remoteKeySet` caches original context
I noticed a non-obvious "bug" in our codebase when using the go-oidc package to verify JWTs. It wasn't expected to me or documented in `oidc.NewProvider` that the supplied context *must never be canceled* if you are going to re-use the resulting `Provider`, though on further investigation it is mentioned in the underlying comments for `oidc.NewRemoteKeySet`. This is a little weird because it means that a `Provider` object created in the context of an incoming HTTP server request cannot be cached and re-used safely in future requests unless you explicitly pass in a separate context that will never be canceled (e.g. `context.Background()`. 

My default assumption was to pass in the context from the HTTP request, but HTTP request contexts are canceled after `ServeHTTP` is finished (according to https://pkg.go.dev/net/http#Request.Context). This led to a fairly hard-to-track-down bug in our codebase since we have a runtime cache of `oidc.Provider`s, but sometimes they would get created during one request but wouldn't fetch JWKS until a future request, at which point the fetch would fail with: `failed to verify signature: fetching keys oidc: get keys failed Get "https://<some-oidc-provider>/.well-known/jwks.json": context canceled`

The fact that `remoteKeySet` internally stores the context runs counter to the Golang standard recommendations of never storing contexts in structs, and always using the one passed in (e.g. the guidance given here https://go.dev/blog/context-and-structs and here https://pkg.go.dev/context). In this case, it seems odd/unintuitive that `remoteKeySet` (which actually accepts a context arg in most methods) would use its own stored context `r.ctx` in `updateKeys()` even though `keysFromRemote` (which is the only method that calls `updateKeys()`) takes a context arg and could just pass that in?

Surprisingly it seems almost no one else has run into this; I only found one other issue via google search that seemed similar (https://github.com/dexidp/dex/issues/1255).

Is this an oversight or intentional?

ericchiang:
This causes NewRemoteKeySet not to detach the context, and adds a method to create a verifier with an explicit context:

```
func (*Provider) VerifierContext(context.Context, *Config) *IDTokenVerifier
```

Fixes https://github.com/coreos/go-oidc/issues/339?

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -60,7 +60,7 @@ func newRemoteKeySet(ctx context.Context, jwksURL string, now func() time.Time) 
 	if now == nil {
 		now = time.Now
 	}
-	return &RemoteKeySet{jwksURL: jwksURL, ctx: cloneContext(ctx), now: now}
+	return &RemoteKeySet{jwksURL: jwksURL, ctx: ctx, now: now}
 }
 
 // RemoteKeySet is a KeySet implementation that validates JSON web tokens against
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -14,6 +14,7 @@ import (
 	"mime"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"golang.org/x/oauth2"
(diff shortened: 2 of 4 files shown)


argenta-quanto (approved):

nic (approved):

nic:
LGTM

Issue #357, opened 2022-10-31
Add device flow authorization support


sealbro:
Hi, guys!

What do you think about supporting [device flow ](https://curity.io/resources/learn/oauth-device-flow/) authorization?

Configuration `.well-known/openid-configuration` contains `device_authorization_endpoint`
like {{KEYCLOAK_URL}}/realms/{{KEYCLOAK_REALM}}/protocol/openid-connect/auth/device

I think adding URL into `providerJSON/Provider` and a method like [here](https://github.com/rjw57/oauth2device/blob/master/oauth2device.go#L86) for waiting for token

ericchiang:
Hey!

The spec you're referring to (https://www.rfc-editor.org/rfc/rfc8628.html) is an OAuth 2.0 specification. It's not related to OpenID Connect (https://openid.net/developers/specs/).

This issue is likely better for Go's oauth2 repo: https://github.com/golang/oauth2

Going to close out since go-oidc is focused on OpenID Connect, not OAuth 2.0. But please re-open if I've misunderstood.

sealbro:
Ok, thanks!
Issue [here](https://github.com/golang/oauth2/issues/418)

PR #365, opened 2023-02-09
support provider device_authorization_endpoint

cbodonnell:
`DeviceAuthURL` was added to `oauth2.Endpoint` in `golang.org/x/oauth2` `v0.13.0`

This PR adds support for parsing the provider `device_authorization_endpoint` from the OIDC discovery endpoint and populates `DeviceAuthURL` for the `oauth2.Endpoint` returned by `Provider.Endpoint()`.

Related issue: https://github.com/coreos/go-oidc/issues/357

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -94,12 +94,13 @@ func doRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
 
 // Provider represents an OpenID Connect server's configuration.
 type Provider struct {
-	issuer      string
-	authURL     string
-	tokenURL    string
-	userInfoURL string
-	jwksURL     string
-	algorithms  []string
+	issuer        string
+	authURL       string
+	tokenURL      string
+	deviceAuthURL string
+	userInfoURL   string
+	jwksURL       string
+	algorithms    []string
 
 	// Raw claims returned by the server.
 	rawClaims []byte
@@ -128,12 +129,13 @@ func (p *Provider) remoteKeySet() KeySet {
 }
 
 type providerJSON struct {
-	Issuer      string   `json:"issuer"`
-	AuthURL     string   `json:"authorization_endpoint"`
-	TokenURL    string   `json:"token_endpoint"`
-	JWKSURL     string   `json:"jwks_uri"`
-	UserInfoURL string   `json:"userinfo_endpoint"`
-	Algorithms  []string `json:"id_token_signing_alg_values_supported"`
+	Issuer        string   `json:"issuer"`
+	AuthURL       string   `json:"authorization_endpoint"`
+	TokenURL      string   `json:"token_endpoint"`
+	DeviceAuthURL string   `json:"device_authorization_endpoint"`
+	JWKSURL       string   `json:"jwks_uri"`
+	UserInfoURL   string   `json:"userinfo_endpoint"`
+	Algorithms    []string `json:"id_token_signing_alg_values_supported"`
 }
 
 // supportedAlgorithms is a list of algorithms explicitly supported by this
@@ -165,6 +167,9 @@ type ProviderConfig struct {
 	// TokenURL is the endpoint used by the provider to support the OAuth 2.0
 	// token endpoint.
 	TokenURL string
+	// DeviceAuthURL is the endpoint used by the provider to support the OAuth 2.0
+	// device authorization endpoint.
+	DeviceAuthURL string
 	// UserInfoURL is the endpoint used by the provider to support the OpenID
 	// Connect UserInfo flow.
 	//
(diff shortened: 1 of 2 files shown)


somersf (commented):
Device code flow support, and the additional `DeviceAuthURL` field in `Endpoint`  _finally_ merged into `golang.org/x/oauth2` @ `v0.13.0`.

Perhaps this PR could be updated and made ready for review by the package maintainers?

Some mini comments:

The "coreos/cbodonnell" search/replace in `README.md`, `go.mod`, `example/{idtoken,userinfo}/app.go`, and `test` should't be part of the PR.

I'd be inclined to update `go.mod` to depend `golang.org/x/oauth2` `v0.13.0`, and to
update oidc/oidc.go `func (p *Provider) Endpoint() oauth2.Endpoint {` to have it initialize the `DeviceAuthURL` field; that way it's auto-populated in the Endpoint for any existing code using `Provider.Endpoint()` - though I'm not sure if the package maintainers would agree or not...

Thanks!

ericchiang:
I don't see device_authorization_endpoint as a field provided by OpenID Connect discovery https://openid.net/specs/openid-connect-discovery-1_0.html

Is there another spec I should be looking at?

ericchiang:
An to be clear, you can retrieve any additional discovery fields through https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims

So users can already access device_authorization_endpoint if a provider returns it today

somersf:
> I don't see device_authorization_endpoint as a field provided by OpenID Connect discovery https://openid.net/specs/openid-connect-discovery-1_0.html
> 
> Is there another spec I should be looking at?

[RFC-8628: OAuth 2.0 Device Authorization Grant -(section 4)](https://www.rfc-editor.org/rfc/rfc8628.html#section-4) defines `device_authorization_endpoint`, and says it can optionally be included in [RFC-8414: OAuth 2.0 Authorization Server Metadata]( https://www.rfc-editor.org/rfc/rfc8414).  The latter RFC's introduction says: _This specification generalizes the metadata format defined by "OpenID Connect Discovery 1.0" [OpenID.Discovery] in a way that is compatible with OpenID Connect Discovery while being applicable to a wider set of OAuth 2.0 use cases._

So, whilst `device_authorization_endpoint` might not be _specifically_ mentioned in an _OIDC_ specification, it can be included/found in `.well-known/openid-configuration` - for example see the final paragraph of [RFC-8414: OAuth 2.0 Authorization Server Metadata - Section 3](https://www.rfc-editor.org/rfc/rfc8414#section-3) says: _Some OAuth applications will choose to use the well-known URI suffix "openid-configuration".  As described in Section 5, despite the identifier "/.well-known/openid-configuration", appearing to be OpenID specific, its usage in this specification is actually referring to a general OAuth 2.0 feature that is not specific to OpenID Connect._


I also felt that because go-oidc's `Provider.Endpoint()` populates an `oauth2.Endpoint`, that it should ideally populate _all_ of the fields in the `oauth2.Endpoint`.  If go-oidc had instead defined/populated its own Endpoint type, I wouldn't have felt that way, as it would have been something independent of the oauth2 struct / OIDC  spec-defined only.

Hope that helps.

somersf:
> An to be clear, you can retrieve any additional discovery fields through https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims
> 
> So users can already access device_authorization_endpoint if a provider returns it today

Yes, thanks, I was aware that it could be extracted this way.  I've tried to provide some references / perspective as to why I felt it should be filled out by `Endpoint()` (now that it is part of `oauth2.Endpoint`) in my reply to your first comment https://github.com/coreos/go-oidc/pull/365#issuecomment-1777925716

cbodonnell:
Thanks @somersf - I see both sides here since it's technically not part of the OIDC spec, but I also see how it fits in given the change to `oauth2.Endpoint`. I'll rebase this PR and address your comments in https://github.com/coreos/go-oidc/pull/365#pullrequestreview-1690831931 this week and we can go from there.

cbodonnell:
@somersf the PR has been updated.

@ericchiang Let us know if this is something reasonable for this project based on the thoughts above. If not, I'd suggest we add a small snippet to the linked issue with the recommended way to retrieve this information for reference.

somersf (commented):
Thanks for the updates - LGTM.  It's over to the project maintainers for their say though!

ericchiang (approved):
lgtm, can you please squash and resolve conflicts? Also do we need to update our go.mod to support newer versions of oauth2 with the added endpoint field?

cbodonnell:
> lgtm, can you please squash and resolve conflicts? Also do we need to update our go.mod to support newer versions of oauth2 with the added endpoint field?

@ericchiang done! the `go.mod` is already on the latest version `v0.13.0` of `golang.org/x/oauth2` as of https://github.com/coreos/go-oidc/commit/3ebd0c9e4dcfceac8e1b23b1fb5ff8e7b8b61c2a, so we are good there.

cbodonnell:
@ericchiang anything else we need to do for this one based on my last comment?

ericchiang (approved):
Sorry for the delay here! lgtm

PR #368, opened 2023-03-21
Support signed JWKs

jaunola:
OpenID Federation specification has signed JWKs, `signed_jwks_uri` in the [metadata](https://openid.net/specs/openid-connect-federation-1_0.html#name-openid-connect-and-oauth2-m).

Some identity providers already support this for added security.

This PR adds support to provide public key which is used to verify signature in the JWT. The signed JWKs are determined by content type "application/jwk-set+jwt". The provider must be configured manually with the call to `NewSignedRemoteKeySet()`. This is because the OpenID discovery document doesn't include `signed_jwks_uri` at the moment but it's only specified in OpenID Federation.

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -54,7 +54,13 @@ func (s *StaticKeySet) VerifySignature(ctx context.Context, jwt string) ([]byte,
 //
 // The returned KeySet is a long lived verifier that caches keys based on any
 // keys change. Reuse a common remote key set instead of creating new ones as needed.
-func NewRemoteKeySet(ctx context.Context, jwksURL string, jwksSigningKey crypto.PublicKey) *RemoteKeySet {
+func NewRemoteKeySet(ctx context.Context, jwksURL string) *RemoteKeySet {
+	return newRemoteKeySet(ctx, jwksURL, nil, time.Now)
+}
+
+// NewSignedRemoteKeySet supports JWKs wrapped inside JWT as described in
+// https://openid.net/specs/openid-connect-federation-1_0.html#name-openid-connect-and-oauth2-m
+func NewSignedRemoteKeySet(ctx context.Context, jwksURL string, jwksSigningKey crypto.PublicKey) *RemoteKeySet {
 	return newRemoteKeySet(ctx, jwksURL, jwksSigningKey, time.Now)
 }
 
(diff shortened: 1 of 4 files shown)


ericchiang:
Thanks for the PR

You've linked to a draft. Is this endpoint in a published version yet? What providers implement this API? Where do users get the public key to verify the other public keys?

jaunola:
Thank you for a quick reply!

> You've linked to a draft. Is this endpoint in a published version yet?

As far as I know, nope. It's only in the linked draft version of OpenID Federation specification.
 
> What providers implement this API?

The usage of this API has been proposed to be used in the Finnish Trust Network and proposed by the Finnish authority Traficom, see their [document](https://www.kyberturvallisuuskeskus.fi/sites/default/files/media/file/Traficom_S213_2023_OIDC_Profile_for_the_Finnish_Trust_Network_EN.pdf), especially chapter 4.

At least one provider in Finland is already implementing this and deprecating regular `jwks_uri` endpoint. Their documentation and API endpoints for testing can be found at https://github.com/op-developer/Identity-Service-Broker-API

> Where do users get the public key to verify the other public keys?

Public key is provided manually by the service broker and configured to Service Provider's application config.

ericchiang:
That context is super helpful!

We don't support draft versions. So for now, I'm inclined to close this out.

Even if this get out of draft, I'm also confused about the threat model and the federation architecture. There are many parts of that document this package doesn't implement (authority_hints, policy_language_crit, trust chain logic, federation entity configuration). What do we gain by supporting signed_jwks_uri without the other aspects of that doc? How do you rotate the public key used to verify the signed JWKs?

Today you can get this logic by using Provider.Claims() to get the signed_jwks_uri, have your own KeySet implementation, then construct a verifier with NewVerifier:

* https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims
* https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#KeySet
* https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#NewVerifier

I'm happy to take this up with other repos (e.g. @op-developer). For now, I'll recommend implement your own KeySet.

I've opened https://github.com/coreos/go-oidc/issues/370 to track full Federation support. If this becomes a popular setup or solves some of our issues with providers like Azure we can revisit this decision.

PR #375, opened 2023-04-24
oidc: add UserInfoEndpoint returning the discocvered URL

### LINKED ISSUE #373: No way to check if server supports UserInfo

raggi:
This enables users detect if the provider.UserInfo method would fail ahead of time, by checking for the empty string in UserInfoEndpoint.

Fixes #373
Fixes #374

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -275,6 +275,12 @@ func (p *Provider) Endpoint() oauth2.Endpoint {
 	return oauth2.Endpoint{AuthURL: p.authURL, TokenURL: p.tokenURL}
 }
 
+// UserInfoEndpoint returns the OpenID Connect userinfo endpoint for the given
+// provider.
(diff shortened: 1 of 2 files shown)


ericchiang:
Thanks!

PR #377, opened 2023-05-10
oidc: Allows client to read configuration

justledbetter:
This provides a copy of the running configuration to clients after it has been obtained from a Discovery Endpoint.  This is accomplished by adding a Config() method to the oidc.Provider type, which loads a new oidc.ProviderConfig struct with the values the Provider is running with and returns it to the user.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -281,6 +281,19 @@ func (p *Provider) UserInfoEndpoint() string {
 	return p.userInfoURL
 }
 
+// Sends back a generated copy of the Provider's configuration, for clients
+// that might need to know this info from an automated Discovery Endpoint.
+func (p *Provider) Config() ProviderConfig {
+	return ProviderConfig{
+		IssuerURL: p.issuer,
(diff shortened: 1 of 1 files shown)


ericchiang:
Thanks for the PR and sorry for the delay here.

See some previous discussion about this here:

https://github.com/coreos/go-oidc/pull/374#issuecomment-1520925016

The correct way to access this config is through https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims. I don't want to add a second way to get at this data without a very strong reason

justledbetter:
Confirmed - Replaced the bits that I needed with a call that pulled back the claims:

```type openIdClaims struct {
        AuthURL     string `json:"authorization_endpoint"`
        TokenURL    string `json:"token_endpoint"`
        UserInfoURL string `json:"userinfo_endpoint"`
        JwksURL     string `json:"jwks_uri"`
}
```
... And it works great!

Apologies for the spurious PR, and thanks again for implementing such a useful library!

PR #378, opened 2023-05-15
Add EdDSA algorithm support

lritter14:
Allows for providers to use EdDSA algorithm but does not support verifying access tokens.


oidc/oidc.go
```diff
@@ -450,6 +452,8 @@ func (i *IDToken) VerifyAccessToken(accessToken string) error {
 		h = sha512.New384()
 	case RS512, ES512, PS512:
 		h = sha512.New()
+	case EdDSA:
+		return errUnsupportedAlgorithm
```
ericchiang: wouldn't this be sha512.New()?

ericchiang (commented):
Out of curiosity, what providers support this?

Would it be possible to add some tests that actually exercise the signing and verification? E.g. take a look at jwks_test.go verify_test.go

If not, I can take a crack at it

oidc/oidc.go
lritter14: I believe that verifying EdDSA key with ed25519 protocol is a bit more complicated than the pattern that exists here [short explanation](https://security.stackexchange.com/questions/233099/validating-an-ed25519-public-key#:~:text=Yes.,see%20if%20the%20equation%20checks.)

lritter14:
> Out of curiosity, what providers support this?
> 
> Would it be possible to add some tests that actually exercise the signing and verification? E.g. take a look at jwks_test.go verify_test.go
> 
> If not, I can take a crack at it

Here is an example provider that supports EdDSA. I was not originally planning on adding verification functionality as it is more complicated for ed25519 protocol.

oidc/oidc.go
ericchiang: This is only for the at_hash value, it's not doing full verification :)

ericchiang:
Did you mean to hyperlink something in that comment?

I don't think I'd want to merge this if it only implements partial support. I can add verification functionality when I get a sec.

lritter14:
> Did you mean to hyperlink something in that comment?
> 
> I don't think I'd want to merge this if it only implements partial support. I can add verification functionality when I get a sec.

lritter14:
I added some additional tests and took your suggestion for using sha512.New() to verify the hash. I am not super knowledgeable in this space so I am not confident in that change.

P.S. closing the PR was a mistake

ericchiang (approved):
Awesome, thanks for sending the full change. Lgtm :)

ericchiang:
If there are any issues with the at_hash support, we can always fix it up later. I'm not super worried about it.

PR #384, opened 2023-06-25
support email_verified value as numeric bool

awwabi:
Can we support '1' and '0' of email_verified as response from auth callback?


ericchiang:
Sorry, this is pretty clearly off spec. You'll need to open a bug with your provider

PR #395, opened 2023-11-16
Support device auth flow

telekid:
This PR adds support for the oauth Device Authorization Flow to `go-oidc`.

The `oauth2` package added Device Authorization Flow support in version `0.13.01`. In order to use this flow, the `DeviceAuthURL` endpoint must be set on the ProviderConfig. This value is made available during dynamic endpoint discovery via the `device_authorization_endpoint` json key. (See https://www.rfc-editor.org/rfc/rfc8628#section-4).

Tested locally against Auth0.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -99,6 +99,7 @@ type Provider struct {
 	tokenURL    string
 	userInfoURL string
 	jwksURL     string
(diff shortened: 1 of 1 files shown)


ericchiang:
Closed by https://github.com/coreos/go-oidc/pull/365

PR #396, opened 2023-11-23
oidc: add device authorization and end session urls

Adphi:


Adphi:
Ugh... I didn't look at the open PRs... I see there's already #226 for logout and #395 for device.

Sorry for the noise.

PR #405, opened 2023-12-14
Typed errors for the verifier

marcinwyszynski:
This Pull Request adds error types to some of the errors returned but the `IDTokenVerifier.Verify` method.
The reasoning here is that in some cases these errors can return more information than you'd like to present to the caller. If you need to perform some filtering on your end (eg. disclose the presented but not the expected audience), it would be more convenient to do it based on well-defined types.

I added tests to ensure that the presented error messages are backwards compatible, and that returned errors are of the correct type. For that I extended the existing testing framework to allow more introspection on returned errors.

I hope it's OK that I used generics here. Looking at the `go.mod` file it should be supported out of the box but if you prefer me to rewrite it without generics, that's not a biggie either.

Thanks!

### DIFF
--- a/oidc/errors.go
+++ b/oidc/errors.go
@@ -1,0 +1,43 @@
+package oidc
+
+import (
+	"fmt"
+	"time"
+)
+
+// TokenExpiredError indicates that Verify failed because the token was expired. This
+// error does NOT indicate that the token is not also invalid for other reasons. Other
+// checks might have failed if the expiration check had not failed.
+type TokenExpiredError struct {
+	// Expiry is the time when the token expired.
+	Expiry time.Time
+}
+
+func (e *TokenExpiredError) Error() string {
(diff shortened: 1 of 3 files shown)


ericchiang:
Hey @marcinwyszynski, happy holidays

Would you mind taking a look at my comment: https://github.com/coreos/go-oidc/pull/406#pullrequestreview-1785515608

These errors happen before the token signature is validated (as a small optimization), so there's likely a lot of scenarios where the data return in the error isn't safe to use. 

In both cases, package users can do audience and issuer validation themselves using some of the Skip options in the Config: https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Config. Which should cover the cases where users want to programmatically do something based on what audience or issuer is returned.

Am I missing any use case here?

PR #408, opened 2023-12-27
Add support for code_challenge_methods_supported

nabokihms:
This is required if an RP wants to use PKCE, but doesn't know whether the provider supports the code challenge or not.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -94,13 +94,14 @@ func doRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
 
 // Provider represents an OpenID Connect server's configuration.
 type Provider struct {
-	issuer        string
-	authURL       string
-	tokenURL      string
-	deviceAuthURL string
-	userInfoURL   string
-	jwksURL       string
-	algorithms    []string
+	issuer               string
+	authURL              string
+	tokenURL             string
+	deviceAuthURL        string
+	userInfoURL          string
+	jwksURL              string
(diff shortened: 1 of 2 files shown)


ericchiang (commented):
Provider.Claims() already lets you access the OpenID Connect metadata

https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#Provider.Claims

Also, what spec is this from and what providers implement this? I'm not seeing it the discovery spec:

https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

I may be missing something, but the added code doesn't actually appear to be used anywhere through the public APIs? E.g. codeChallengeMethods is assigned but never read or exported. Is this part of some larger change?

nabokihms:
This is from the Rfc 8414 https://www.rfc-editor.org/rfc/rfc8414.html (missed that this is not a part of the oidc spec).
I think claims is what I need, thank you for pointing me.

ericchiang:
Thanks! For what it's worth, I think adding support for supplementary specs is great. I'd just want to figure out if there's any additional API we can present more than parsing the string.

E.g. what we do for the [at_hash support](https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#IDToken.VerifyAccessToken) or device authorization https://github.com/coreos/go-oidc/pull/365

nabokihms:
Closing since we solved the problem the way ericchiang suggested.

Issue #402, opened 2023-12-08
Inherit Provider http client settings in calls


p53:
Right now when initializing Provider we are passing to it all settings in context, but when doing Token() calls or e.g. Userinfo calls these settings are totally ignored so we have to do double work and again and again set settings before each call which is quite error prone (you would not expect that you have to do it before each call) + tedious, it would be great if these calls by default would take client from provider and in case client is provided directly in context passed to method call take those

e.g. We are setting up provider but those settings are ignored in calls

```
restyClient := client.RestyClient()
	restyClient.SetTimeout(r.Config.OpenIDProviderTimeout)
	restyClient.SetTLSClientConfig(
		&tls.Config{
			InsecureSkipVerify: r.Config.SkipOpenIDProviderTLSVerify,
		},
	)

	if r.Config.OpenIDProviderProxy != "" {
		restyClient.SetProxy(r.Config.OpenIDProviderProxy)
	}

	ctx := oidc3.ClientContext(context.Background(), restyClient.GetClient())
	var provider *oidc3.Provider
	var err error

	operation := func() error {
		provider, err = oidc3.NewProvider(ctx, r.Config.DiscoveryURL)
```

ericchiang:
Hey @p53 

The clients and options are passed through contexts to match Go's OAuth 2.0 package, which uses the same context key/value strategy: https://pkg.go.dev/golang.org/x/oauth2#pkg-variables. If you use go-oidc with the oauth2 package (which was the main goal of go-oidc), you have to remember to construct the context anyway for the oauth2 package.

Design for how the contexts should be used has generally been a really big pain point: https://github.com/coreos/go-oidc/issues/214.

In hindsight I wouldn't use the context keys for something like HTTP Client configuration. But for now, I don't think I want to be silently inferring values or anything clever. So I'm inclined to add some documentation or examples rather than change the context logic (again).

p53:
from point of maintainer i understand, usually making too many clever things leads to bugs, but from point of user sad, it is pain for me, already made several times mistake because of forgetting about this, adding any values to context was not good idea anyway, usually misused as always when you make some "global" var available, ok thx anyway :)

PR #413, opened 2024-02-22
Add access to Provider client from the provider

nathenapse:
This is helpful when we want to call Exchange with the client we provided to the Provider

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -289,6 +289,12 @@ func (p *Provider) UserInfoEndpoint() string {
 	return p.userInfoURL
 }
 
+// ClientContext returns a new context that carries the given HTTP client with
+// provider.
+func (p *Provider) ClientContext(ctx context.Context) context.Context {
+	return ClientContext(ctx, p.client)
+}
+
 // UserInfo represents the OpenID Connect userinfo claims.
 type UserInfo struct {
 	Subject       string `json:"sub"`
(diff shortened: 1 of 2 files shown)


ericchiang:
I don't see a strong reason the Provider struct should allow accessing it's internal fields like this. If you need to use the same http.Client as the Provider, use the http.Client you passed to the Provider in the first place?

nathenapse:
Let me explain my reasoning behind it. When passing the `*http.client` to the `NewProvider()` I expected it to also affect the `Exchange()` but that was not the case. So if we also need to pass the `*http.client` to the `Exchange()` it would be better if we override the `Exchange()` method to pass the `http.client` found in `Provider`. Or have a way to access it and pass it to `Exchange()`. Clearly I chose the second method but the first option is also something I can work on. What do you think?

ericchiang:
Ah. Would you take a look at my comment here https://github.com/coreos/go-oidc/issues/402#issuecomment-1847505771. That goes over some other unexpected behavior with passing config through contexts.

This is unfortunate, but I think the answer remains that you should hold on to the http.Client configuration and reuse it between calls. Having multiple ways to get at the client doesn't seem like it's adding much at the cost of more API surface

nathenapse:
That means if we have a custom `*http.client` we have to remember to pass it everytime? 
Wouldn't that be prone to errors? and sometimes tls configurations are production specific.

ericchiang:
This is the case for all values passed as context keys. If you use a context, it needs to be configured correctly before passing it to your method. E.g. if you use golang.org/x/oauth2, you also need to do the same.

Again, as I mentioned in #402, if we're going to reduce confusion, we should allow HTTP client configuration without using contexts. I don't think this PR is the correct way to improve things, so going to close out

PR #416, opened 2024-03-18
fix error message parse jws

k-omotani:


ericchiang:
Thanks, but I don't think this is a point of confusion. Going to leave this as is

PR #425, opened 2024-04-04
Allow Header Injection

spjmurray:
Use case is for tracing/performance profiling through a distributed system, in particular the W3C trace context standard used by open-telemetry.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -84,10 +84,34 @@ func InsecureIssuerURLContext(ctx context.Context, issuerURL string) context.Con
 	return context.WithValue(ctx, issuerURLKey, issuerURL)
 }
 
+// HeaderInjectionFunc allows a client to inject headers into requests.
+type HeaderInjectFunc func(http.Header)
+
+type headerInjectContextKey int
+
+var headerInjectKey headerInjectContextKey
+
+// HeaderInjectClientContext allows a callback to be registered that can inject headers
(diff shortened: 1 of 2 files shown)


spjmurray:
Open for discussion as always!  Using it is as simple as 

```go
import (
       "go.opentelemetry.io/otel"
       "go.opentelemetry.io/otel/propagation"
)

func foo(ctx context.Context) {
       ctx = oidc.HeaderInjectClientContext(ctx, func(h http.Header) {
               otel.GetTextMapPropagator().Inject(ctx, propagation.HeaderCarrier(h))
       })
       // Provider() or Userinfo() call
}
```

And it works quite nicely!

Screenshot from 2024-04-04 14-25-57

spjmurray:
Bump.  Someone, anyone?  I ask not much, I promise thee!  I even offer beer....

ericchiang:
Thanks for the ping, sorry about that!

In my mind this is what accepting a custom http.Client with a custom transport is for. Go's oauth2 package does something similar to set the bearer token auth header.

Does that work for you? I imagine that strategy would avoid having to make change to other libraries as well.

Do you have examples of how other packages are supporting this?

spjmurray:
Ah, that's how I missed it, it's not a callback but a receiver, my bad!  Does make the code somewhat less legible, but that's just Go for you 😿

spjmurray:
Okay, finally got a chance to test it out, and it works!  Cheers

PR #438, opened 2024-10-29
fix: wrap error messages in fmt.Errorf

jamescoombs-soda:
use %w instead of %v when nesting error messages so they can be properly unwrapped by upstream callers.

I've run into a problem where I'm unable to use errors.Is() in my tooling to help properly respond to errors coming out of oidc. The primary use case is checking for context.Canceled, but I'm sure there are others.

Note: I believe %w was introduced in Go version 1.13, so this change isn't friendly to users who use a 5 year old Go version but keep their packages up to date. In any case, it's possible this package already requires a higher Go version than that.

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -29,7 +29,7 @@ func (s *StaticKeySet) VerifySignature(ctx context.Context, jwt string) ([]byte,
 	// any algorithm.
 	jws, err := jose.ParseSigned(jwt, allAlgs)
 	if err != nil {
-		return nil, fmt.Errorf("parsing jwt: %v", err)
+		return nil, fmt.Errorf("parsing jwt: %w", err)
 	}
 	for _, pub := range s.PublicKeys {
 		switch pub.(type) {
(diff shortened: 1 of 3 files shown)


ericchiang:
Thanks for the PR!

I'm against using wrapped errors overly-broadly like this without making it part of the API contract. At the very least, there needs to be an associated test, since users might depend on a method returning a specific error type.

You mentioned checking for context.Canceled. Do you have an example? Particularly with the new "Cause" API, there are other ways of handling context cancelation.

https://pkg.go.dev/context#Cause

Do you have a sense of what the godoc changes would be before we move forward with the code itself?

jamescoombs-soda:
I'll close this. I've forgotten about it, it's stale, I've found another way to solve my problem, and while I think it's a good change I don't have the capacity to run through the extra steps you'd want to see.

I will say that I personally don't think this data needs to be added to the API contract. I think adding it to the contract (and having the test) is the signal to users that the specifics of the functionality can be relied on. Simply returning data in a more idiomatic way that persists more data in it's original format is valuable, and doesn't necessarily imply to the user it's data they should use. And, for what it's worth, there's other places in this package that use %w as I did here.

Issue #422, opened 2024-03-22
Error in parsing remoteKeySet endpoint


s1ovac:
Could be error inside parsing remoteKeySet
After library update to v3.10.0 from 3.9.0 got this error in parsing remote key.
oidc: invalid userinfo jwt signature fetching keys oidc: failed to decode keys: got Content-Type = application/json, but could not unmarshal as JSON: illegal base64 data at input byte 342 {"keys":[{"kid":"unique-key-identifier","use":"sig","alg":"RS256","kty":"RSA","e":"AQAB","n":"qssJQDlZeVfdJLTZP_pJl_1WfXXFwOc48L_sWnKloogwxTZ9SBIdn5k9mzSHQ7qsBCELO5imdZU5ubCxfN_DTcILFD3uVw1_H2zDlM2JdxLXAL6ao1pcTAQ3jEOYC8z1m6Nf8gEhnmQenQFliJM1Rma8bfs9-OOrVU2SdMiyBSFZ0g-KuUYIdAHrtcnmJhu9QHpqqvBY6VzyKLENShbWf1lM_fBqDhatdieiNgCZhRvxvfM4ZPrIKF5DCIw34ROp4ZoSkRBA6LVINDTHn2NFQ9_CLAyU8sLKJFEQkFgcOn85mAUD46rc1Vlw95Uc_JiZfPYe14QybQutQmIUsd5SmQ=="}]}

ericchiang:
Hey @s1ovac, it looks like this is was caused by upgrading https://github.com/go-jose/go-jose #421 , which does the serialization for these types. I don't believe the key you posted is a valid key according to the spec? The JOSE specs use non-padded base64. 

https://www.rfc-editor.org/rfc/rfc7515.html#section-2

> Base64 encoding using the URL- and filename-safe character set
      defined in [Section 5 of RFC 4648](https://www.rfc-editor.org/rfc/rfc4648#section-5) [[RFC4648](https://www.rfc-editor.org/rfc/rfc4648)], with all trailing '='
      characters omitted

See one of the examples from the JWK spec: https://www.rfc-editor.org/rfc/rfc7517.html#page-25

Or https://www.googleapis.com/oauth2/v3/certs

```
{
  "keys": [
    {
      "kty": "RSA",
      "use": "sig",
      "e": "AQAB",
[...]

Issue #230, opened 2020-01-26
Cannot import Go module


misberner:
Super happy to see Go modules support came to this project with the most recent release! Unfortunately, it looks like the major version of 2 requires the module to declare its path as `github.com/coreos/go-oidc/v2`, as I'm seeing:
```
require github.com/coreos/go-oidc: version "v2.2.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

Happy to send out a PR, but wondered whether that should just migrate all package names to `.../v2`, or if this needs to be bumped to `v3` as the package name change itself would be backwards-incompatible.

NOTE: I saw that there was some discussion around possibly making this a `/v2` module on the PR, but I think the way Go modules works, this is actually non-optional. Even if I append a `+incompatible` I get this error:
```
verifying github.com/coreos/go-oidc@v2.2.0+incompatible/go.mod: github.com/coreos/go-oidc@v2.2.0+incompatible/go.mod: reading https://sum.golang.org/lookup/github.com/coreos/go-oidc@v2.2.0+incompatible: 410 Gone
```

ericchiang:
Thanks for opening this issue. I'm going to go ahead an revert the go.mod for v2 and we can discuss a v3 :)

ericchiang:
I field #232 for some thoughts on a v3

PR #447, opened 2025-01-22
Switch to maintained go-jose library

liggitt:
This allows consumers of the v2 stream to pick up the fix for https://pkg.go.dev/vuln/GO-2024-2631

Switching to v3 changes token generation (due to changes in go-jose) in ways that make it difficult to update without visible behavior changes. Without this, Kubernetes may have to fork the v2 branch of this dependency to update to go-jose. cc @dims for visibility (xref https://github.com/kubernetes/kubernetes/pull/129732#issuecomment-2604924444)

The TMP commits would be dropped and are only there to demonstrate the effective diff in the dependency.

After review, the TMP commits would be dropped and this could be merged and tagged as v2.3.0, etc

* The `Switch to maintained gopkg.in/go-jose/go-jose.v2 library` commit is the only change intended to be merged (only switches import paths for the go-jose library)
* The `TMP: revendor to show effective dependency diff` commit shows the actual dependency change

### DIFF
--- a/jwks.go
+++ b/jwks.go
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/pquerna/cachecontrol"
-	jose "gopkg.in/square/go-jose.v2"
+	jose "gopkg.in/go-jose/go-jose.v2"
 )
 
 // keysExpiryDelta is the allowed clock skew between a client and the OpenID Connect
--- a/jwks_test.go
+++ b/jwks_test.go
@@ -14,7 +14,7 @@ import (
 	"testing"
 	"time"
 
-	jose "gopkg.in/square/go-jose.v2"
+	jose "gopkg.in/go-jose/go-jose.v2"
 )
 
 type keyServer struct {
--- a/oidc.go
+++ b/oidc.go
@@ -18,7 +18,7 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
-	jose "gopkg.in/square/go-jose.v2"
+	jose "gopkg.in/go-jose/go-jose.v2"
 )
 
 const (
--- a/verify.go
+++ b/verify.go
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
-	jose "gopkg.in/square/go-jose.v2"
+	jose "gopkg.in/go-jose/go-jose.v2"
 )
 
 const (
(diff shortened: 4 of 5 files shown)


dims:
thanks @liggitt

liggitt:
cc @ericchiang

ericchiang:
Thanks! let me fix up the CI first, then will merge

For the breaking changes

> Switching to v3 changes token generation (due to changes in go-jose) in ways that make it difficult to update without visible behavior changes

I'm aware of stricter parsing of base64 encodings (https://github.com/coreos/go-oidc/issues/422#issuecomment-2015376190), but is there anything else you're seeing?

dims:
@ericchiang thanks for the quick response! we saw this as well - payload changes needed in unit tests - https://github.com/kubernetes/kubernetes/pull/123649#discussion_r1510166809

ericchiang:
Thanks, will take a look when I get a change.

If you rebase over https://github.com/coreos/go-oidc/pull/448 you should be good to go, and I can tag a release

go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/coreos/go-oidc
```
ericchiang: Last time we did this we broke imports of this branch.

https://github.com/coreos/go-oidc/issues/230
https://github.com/coreos/go-oidc/pull/231

I don't know if that's changed in the upstream Go tooling since then.

ericchiang:
For my own reference, the changes to the kubernetes tests are based on their usage of go-jose's JWT library to generate tokens. go-oidc doesn't support creation of ID Tokens, and handles these encoding differences fine, so there shouldn't be any observable changes to other go-oidc users.

https://github.com/kubernetes/kubernetes/pull/129732#issuecomment-2607891117

go.mod
liggitt: all the go.mod and vendor bits are in TMP commits I would drop before we merge this, I just included those so we could actually see the effective change in the dependency code the square → go-jose change would make

if you're done looking at the vendor change, I'll drop those commits out completely

liggitt: I agree we don't want to add go.mod to the v2 branch

liggitt: done, dropping TMP commits

ericchiang:
Tagged https://github.com/coreos/go-oidc/releases/tag/v2.3.0

dims:
thanks @ericchiang !

liggitt:
thanks so much, appreciate it

PR #461, opened 2025-07-12
fix confusing error message

vladmsnk:
When I try to connect to Oauth provider with wrong url I got wrong error message. In example I use localhost:4444 instead of 127.0.0.1. 

Error message:
`error="failed to create OAuth service: failed to create OIDC provider: oidc: issuer did not match the issuer returned by provider, expected \"http://localhost:4444\" got \"http://127.0.0.1:4444\""`

But should be:
`error="failed to create OAuth service: failed to create OIDC provider: oidc: issuer did not match the issuer returned by provider, expected \"http://127.0.0.1:4444\" got \"http://localhost:4444\""`

When i change host to http://127.0.0.1:4444, everything works correctly and got no errors

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -162,7 +162,7 @@ var supportedAlgorithms = map[string]bool{
 // parsing.
 //
 //	// Directly fetch the metadata document.
-// 	resp, err := http.Get("https://login.example.com/custom-metadata-path")
+//	resp, err := http.Get("https://login.example.com/custom-metadata-path")
 //	if err != nil {
 //		// ...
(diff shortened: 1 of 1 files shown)


vladmsnk:
@ericchiang Hi! could you check this PR please?

ericchiang:
Hey! Sorry, but seems like the correct ordering. If the URL you provide to this library is http://localhost:4444, and the server returns something different, then the library expects localhost but got a different value.

Maybe "configured to expect localhost, but server returned 127.0.0.1" would be clearer?

PR #462, opened 2025-07-14
feat: add typed error for issuer mismatch

mstott2:
Similar to https://github.com/coreos/go-oidc/pull/342. Allows programs to better distinguish and act on issuer mismatch errors. 

When working with the OIDC client, I want to check for different `oidc.NewProvider` failures for more intelligent error handling and determining response codes. For issuer mismatch errors, I currently need to rely on string matching to differentiate these from other cases, but it would be preferable to have a typed error for better comparison checks and more easily access the expected and actual issuer values.

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -40,6 +40,16 @@ var (
 	errInvalidAtHash = errors.New("access token hash does not match value in ID token")
 )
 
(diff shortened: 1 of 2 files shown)


ericchiang:
Fixed by https://github.com/coreos/go-oidc/pull/493

PR #464, opened 2025-07-30
oidc: verify the ID Token's signature before processing claims

ericchiang:
This change updates the verification logic of this library to first validate the ID Token instead of parsing claims. This hopefully makes it harder for a malicious client to provide an invalid token for validation that's crafted to cause this package to over-allocate memory. See the associated bug and CVE-2025-27144.

Fixes https://github.com/coreos/go-oidc/pull/463

### DIFF
--- a/oidc/jwks_test.go
+++ b/oidc/jwks_test.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
Can you publish an advisory at https://github.com/coreos/go-oidc/security ?

PR #465, opened 2025-08-02
oidc: verify signature before parsing token

ericchiang:
This change updates the verification logic of this library to always verify the signature of the token before validating the payload. See associated issue.

https://github.com/coreos/go-oidc/pull/464

### DIFF
--- a/verify.go
+++ b/verify.go
@@ -1,11 +1,9 @@
 package oidc
 
 import (
-	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


verify.go
```diff
@@ -211,12 +209,29 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
...
+	supportedSigAlgs := v.config.SupportedSigningAlgs
+	if len(supportedSigAlgs) == 0 {
+		supportedSigAlgs = []string{RS256}
+	}
+
+	if !contains(supportedSigAlgs, sig.Header.Algorithm) {
+		return nil, fmt.Errorf("oidc: id token signed with unsupported algorithm, expected %q got %q", supportedSigAlgs, sig.Header.Algorithm)
+	}
+
+	payload, err := v.keySet.VerifySignature(ctx, rawIDToken)
```
liggitt, line 230: how did moving this first help the DOS? this still calls:

```
VerifySignature → jose.ParseSigned → parseSignedCompact
```
which still calls `parts := strings.Split(input, ".")`

If this was trying to catch spurious tokens with lots of `.` characters, something like the count check in https://go.dev/cl/652155 would seem more effective

ericchiang: Hey! The v3 version of this package depends on a go-jose version that doesn't do that. 

https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/go.mod#L8
https://github.com/go-jose/go-jose/blob/v4.1.3/jws.go#L367

For v2, maybe we should do our own check before this to throw out invalid ID tokens early

liggitt: > For v2, maybe we should do our own check before this to throw out invalid ID tokens early

Yeah, that's exactly what I'd recommend

liggitt: (also, sorry for the random 6-month-later drive-by ... I just got here while reviewing https://github.com/kubernetes/kubernetes/pull/136162#discussion_r2687404738)

ericchiang: Lol, no worries. Appreciate you actually checking my work.

Sent the change and tagged a release:
* https://github.com/coreos/go-oidc/pull/473
* https://github.com/coreos/go-oidc/releases/tag/v2.5.0

PR #466, opened 2025-08-08
refactor: Remove unused time injection from RemoteKeySet

ponimas:
Remove the time injection parameter and now function field from RemoteKeySet as they were not being used in the codebase. This simplifies the constructor signature and reduces unnecessary complexity in the JWKS key set implementation. The time package import was also removed as it's no longer needed.


ericchiang:
thanks!

PR #467, opened 2025-10-03
bump go to 1.24, remove 1.23 support, bump go-jose dependency, remove x/net dependency

wardviaene:
This PR bumps the go-jose dependency to the latest version. It sets the Go version to 1.24, and removes 1.23 in the GitHub workflow file. Also the golang.org/x/net dependency is removed.

The goal is to remove external dependencies of go-oidc, including go-spew, which is not maintained anymore and rejected by some projects, like kubernetes (https://github.com/kubernetes/kubernetes/issues/103942). Removing this dependency will also help to bump go-oidc in kubernetes (https://github.com/kubernetes/kubernetes/issues/123252).


ericchiang:
thanks! I'll tag a release once this is in.

ericchiang:
There you go: https://github.com/coreos/go-oidc/releases/tag/v3.16.0

PR #470, opened 2025-12-01
refactor: align struct fields  with OpenID Connect specification

HuckOps:
This change adds Name, Picture, GivenName, FamilyName in the UserInfo structure to solve the problem of not being able to obtain user information in some scenarios


ericchiang:
Thanks for the PR. Historically, we haven't added every new field in the spec since many of them aren't implemented in practice. What providers return these?

HuckOps:
I found that these fields exist when using keycloak and casdoor.
image
And I think the name field is a key field that needs to be included in the response.

Issue #355, opened 2022-10-21
Verify azp claim 


manjinder-singh-tech:
Is there a specific reason, verifier does not support verifying `azp` claim? It could be possible for some scenarios that client is explicitly required to verify that `azp` claim matches ClientID. Can we add support for verifying that by any chance, it does not have to be enforced and can be optional (via verifier config) to satisfy certain use cases for say custom providers.

ericchiang:
There's not a strong reason. I've only seen azp used in relatively niche scenarios where you still want to verify the audience rather than authorizing party.

Do you have a use case in mind? What's the desired security outcome?

manjinder-singh-tech:
We have a custom Identity Provider, that for some reason supports other entities as ‘aud’ rather than ClientID, and the authorized party sends ClientID as part of ‘azp’ in the token, and requires us to validate that instead of aud. 

I understand that perhaps this not a very common use case, but hoping it can still be beneficial to others. If we can provide optional config attribute to validate ‘azp’ to match ClientID.

ericchiang:
Today you can pull the 'azp' claim out of the IDToken with the Claims() method and do your validation on that.

https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#IDToken.Claims

It does appear that the azp claim is supposed to be validated after the token endpoint call

"If an azp (authorized party) Claim is present, the Client SHOULD verify that its client_id is the Claim Value."

https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

The problem is that this package can be used by both the server requesting the token and other service that might want to validate the token later. For the former, you want to verify 'azp' matches your client_id. For the latter, you'd want to validate 'aud'.

So yeah, maybe we should have an "authorizing party" validation option?

manjinder-singh-tech:
Makes sense i believe. We can make it optional to satisfy both server and client implementations, and the verifier config can be adjusted accordingly.
Aldo, thanks for suggesting the Claims method.

ericchiang:
Another thought is that we could have a method like:

```
func (v *IDTokenVerifier) VerifyToken(ctx context.Context, token *oauth2.Token) (*IDToken, error)
```

That did the right thing with the authorizing party claim. Might be a bit subtle though... Let me think about this a little

trixpan:
@ericchiang the spec states 

> 4. If an azp (authorized party) Claim is present, the Client SHOULD verify that its client_id is the Claim Value.

[...]

PR #474, opened 2026-02-09
oidc: optional azp (authorized party) validation

1seal:
this adds an opt-in verifier option to validate the "azp" (authorized party) claim against Config.ClientID.

when Config.VerifyAuthorizedParty is enabled:
- if aud has multiple values, azp must be present and equal to ClientID
- if azp is present, it must be equal to ClientID

default behavior is unchanged.

refs: #355, oidc core id token validation https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

### DIFF
--- a/oidc/oidc.go
+++ b/oidc/oidc.go
@@ -501,16 +501,17 @@ func (i *IDToken) VerifyAccessToken(accessToken string) error {
 }
 
 type idToken struct {
-	Issuer       string                 `json:"iss"`
-	Subject      string                 `json:"sub"`
-	Audience     audience               `json:"aud"`
-	Expiry       jsonTime               `json:"exp"`
-	IssuedAt     jsonTime               `json:"iat"`
-	NotBefore    *jsonTime              `json:"nbf"`
-	Nonce        string                 `json:"nonce"`
-	AtHash       string                 `json:"at_hash"`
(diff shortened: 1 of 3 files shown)


ericchiang:
Linking my comment in https://github.com/coreos/go-oidc/issues/355#issuecomment-3874581435.

I don't plan on merging this PR and won't add additional API surface for azp or multi-audience behavior without a real world provider that this matters for.

1seal:
thanks for the detailed context, that makes sense.

i agree the current option in #474 is too opinionated: it bakes in the “post-token-endpoint client” interpretation where azp must bind to the configured ClientID, but go-oidc is also used in “third-party verifier” scenarios (e.g. kubernetes) where azp being different from a particular relying party identifier can be intentional.

given i don’t have a concrete real-world provider interaction to point to, and since callers can already implement their own azp policy today via IDToken.Claims, i’m going to close this pr to avoid adding api surface without a strong motivating case.

if you’d accept a small docs-only change, i can send a separate pr that documents the trade-off and shows a minimal example of extracting/validating azp (and when it might be appropriate).,

1seal:
closing to avoid adding api surface without a real-world motivating provider/interaction; happy to revisit with a concrete repro or send a docs-only pr if desired.

PR #475, opened 2026-03-30
Added support for non-standard issuer (e.g fusionauth.io) that do not…

digitive:
This PR fixed issue that some IDP (e.g. fusionauth.io) their issuer do not have `https://` prefix.


ericchiang:
Please report this as a bug to fusionauth.io. There are numerous places in the spec where the issuer URL is defined as a URL with a scheme

https://openid.net/specs/openid-connect-core-1_0-final.html#Terminology

digitive:
Agreed. Thanks for the check.

PR #476, opened 2026-04-07
Bump github.com/go-jose/go-jose/v4 to v4.1.4

rhafer:


ericchiang:
Hey @rhafer. go-oidc doesn't require any newer features from go-jose, so I don't see a reason to bump the dependency and require all of our consumers do so too. If you want to rely on a newer version of go-jose, it's easy enough to do so in your own project's `go.mod`.

ericchiang:
Ah, I see https://github.com/coreos/go-oidc/security/dependabot/24. Let me go ahead and use this to configure dependabot for the repo.

ericchiang:
Should be fixed by https://github.com/coreos/go-oidc/pull/479!

PR #482, opened 2026-06-08
feat(oidc): rate-limit remote JWKS fetches every 30s

BaptisteLalanne:
## Description:

This is a suggestion to limit the risk of flooding a remote with requests to get keys for a JWT that is using a random/unknown KID.

Currently an unknown KID triggers a fetch to the remote to update the cached keys which respects the openid specs
https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys

> Keys can be rolled over by periodically adding new keys to the JWK Set at the jwks_uri location. The signer can begin using a new key at its discretion and signals the change to the verifier using the kid value. The verifier knows to go back to the jwks_uri location to re-retrieve the keys when it sees an unfamiliar kid value

Yet, in case of a unexpected JWTs hitting the server with a random KID, it might lead to elevated fetches on the remote. 

For JWKS setup, 30s is expected to be really fair. New keys are usually not expected to be created that often nor used right away.
- For server **without** any unexpected KID check:
  - a new KID will trigger an update of the cache, 0 errors
- For a server **with** unexpected KID check:
  - a new KID introduced and directly used (edge case) might create errors for up to 30s if the cache was wrongly refreshed on an invalid JWKS

Note: jose JS implementation does use 30s as well by default https://github.com/panva/jose/blob/a90ad6e683efb74539235b2363d1f3bda3ab6ce1/src/jwks/remote.ts#L261-L265 

## Changes: 
Prevent JWTs with arbitrary kid values from flooding the remote JWKS endpoint. keysFromRemote now skips the HTTP GET and returns cached keys if a succesful fetch was initiated within the last 30 seconds.

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net/http"
 	"sync"
+	"time"
 
 	jose "github.com/go-jose/go-jose/v4"
 )
@@ -72,6 +73,11 @@ func newRemoteKeySet(ctx context.Context, jwksURL string) *RemoteKeySet {
 	}
 }
 
+// remoteKeyFetchInterval is the minimum time between remote JWKS fetches.
+// It caps how often unknown key IDs can trigger HTTP GETs, preventing remote
+// endpoint flooding from JWTs carrying arbitrary kid values.
+const remoteKeyFetchInterval = 30 * time.Second
+
 // RemoteKeySet is a KeySet implementation that validates JSON web tokens against
 // a jwks_uri endpoint.
 type RemoteKeySet struct {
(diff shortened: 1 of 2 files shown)


ericchiang:
Thanks!

This package already dedupes key fetches, so at most a malicious request can only result in a single fetch at any given time.

You'll also end up with a ton of 500s which you should likely rate limit at a higher level. If you're looking to overwhelm the server you can already go directly to the upstream

Unless this is causing real impact, I'm going to close for now.

BaptisteLalanne:
@ericchiang Thanks for having a look  🙇 

The package does have inflight request handler yes but it was really for the scenario of clients continuously hitting the server with an invalid KID due to X reasons (ex: key removed from jwks_uri by mistake still used by clients, stale DNS record hitting the wrong service...) and not just for a one time request.

The proposal really wasn't purposely for malicious protection but rather for upstream protection in case of incident

PR #485, opened 2026-06-15
fix: Key refresh should set `no-cache` to get most up to date keys

### LINKED ISSUE #484: Refresh doesn't send `Cache-Control: no-cache` header.
### Problem
Currently the call to the jwks endpoint doesn't set any `Cache-Control` headers, which means that potential proxies in between might serve cached content that is no longer up to date (for example, since a new `kid` was added).

### Solution
The `updateKeys` function should send a `no-cache` header to ensure that the keys are truly refreshed and not just stale keys fetched from a proxy in between.

https://github.com/coreos/go-oidc/blob/v3/oidc/jwks.go#L236-L263

Yanni8:
Currently, the HTTP client fetches the keys without a `Cache-Control` header. This can lead to potential issues because, with the current logic, we always attempt to revalidate when we encounter an unknown `kid`. Therefore, we want to ensure that we retrieve the most up-to-date keys from the OIDC server.

For this reason, we should include a `Cache-Control: no-cache` header in the request to ensure that the keys are revalidated and that we do not receive a stale key set from an intermediary proxy.

Fixes #484

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -238,6 +238,7 @@ func (r *RemoteKeySet) updateKeys() ([]jose.JSONWebKey, error) {
 	if err != nil {
 		return nil, fmt.Errorf("oidc: can't create request: %v", err)
 	}
+	req.Header.Set("Cache-Control", "no-cache")
 
 	resp, err := doRequest(r.ctx, req)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ericchiang (approved):
thanks!

PR #497, opened 2026-08-02
example: use PKCE in the authorization code flow

### LINKED ISSUE #496: Add PKCE to examples, link from README and godoc
There was a write up on the Go subreddit and this was something that was a bit confusing

https://www.reddit.com/r/golang/comments/1v7lxmy/implementing_oidc_with_entra_id/

ryandiginomad:
## Problem

The three programs under `example/` run the authorization code flow without PKCE, and neither the README nor the package documentation mentions it anywhere. Someone starting from an example therefore gets a flow with no protection against authorization code injection — which [RFC 9700 §2.1.1](https://datatracker.ietf.org/doc/html/rfc9700#section-2.1.1) requires of public clients and recommends for confidential ones — and some providers reject an exchange that omits it.

`golang.org/x/oauth2` already provides `GenerateVerifier`, `S256ChallengeOption`, and `VerifierOption`, and this module depends on v0.36.0, so nothing new is pulled in.

Closes #496

## Changes

Two commits:

1. **`example:`** — all three examples generate a verifier per authorization request, send its S256 challenge on the redirect, and replay the verifier on the exchange. The verifier rides on the existing `setCallbackCookie` helper, exactly the way `state` and `nonce` already cross the redirect, so this mirrors what each example does rather than introducing a second mechanism. (Named `codeVerifier` because `verifier` is already taken by the ID Token verifier in two of the three.)

2. **`README, oidc:`** — document the verifier/challenge pair in the README snippets and in `oidc/doc.go`, and add a README section linking the three programs. The README previously never pointed at `example/` at all.

## Testing

Ran each example against a live discovery document and captured the redirect:

- `example/idtoken` and `example/userinfo` against `accounts.google.com`
- `example/logout` against `demo.duendesoftware.com`, since Google publishes no `end_session_endpoint`

All three emit `code_challenge` and `code_challenge_method=S256` on the authorization URL and set an `HttpOnly` `code_verifier` cookie, with `state`/`nonce` unaffected. I checked the pair is cryptographically consistent rather than merely present — from the `idtoken` run:

```
code_verifier               = _QJdctEKTn4qLxe7k70EEn8b5v4akgLusDWJDRJM0WY
code_challenge              = v6QQksnXv3_L_u-v4Onc_8dZheRhZRdFi8KfJVR-LzY
BASE64URL(SHA256(verifier)) = v6QQksnXv3_L_u-v4Onc_8dZheRhZRdFi8KfJVR-LzY   ← match
```

I also drove the same call pattern through a stub token endpoint to confirm the token request carries a `code_verifier` equal to the original, that the plaintext verifier never leaks onto the front channel, and that dropping the options sends no PKCE parameters at all. That was scratch verification and is not in this PR, since `example/` carries no tests.

`gofmt -l .` is clean; `go build ./...`, `go vet ./...`, and `go test ./...` pass.

## Open to feedback

- **Scope reading.** I read "add PKCE to examples" as "make the examples do it" rather than "add a separate PKCE example", since a standalone one would imply PKCE is an opt-in variant. Happy to split it out instead if you meant the latter.
- **Touching the README/godoc snippets.** I updated those so they don't contradict the examples. If you'd rather the prose stayed as it was and only gained the link, that's a smaller diff and I'll trim it.
- **The linked thread.** I could not read the Reddit post (Reddit blocks automated fetching), so this is written from the repo state — no PKCE in code, docs, or examples — rather than from the specific confusion described there. If the thread was pointing at something narrower, tell me and I'll retarget.

### 🤖 AI assistance

Written with Claude Code. I reviewed every line and ran the verification above myself. One correction worth surfacing: the draft first claimed RFC 9700 requires PKCE for *every* authorization code exchange. Checking the RFC showed that is only true for public clients — for confidential ones it is RECOMMENDED, and confidential OpenID Connect clients MAY use `nonce` instead — so the wording now reflects that distinction. Since these examples are confidential clients that already send a nonce, the accurate framing is "recommended and widely required in practice", not "mandated".

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,13 +80,26 @@ oauth2Config := oauth2.Config{
 idTokenVerifier := provider.Verifier(&oidc.Config{ClientID: clientID})
 ```
 
-OAuth2 redirects are unchanged.
+OAuth2 redirects are unchanged, except for PKCE ([RFC 7636][pkce]). The [OAuth 2.0
+Security Best Current Practice][oauth2-bcp] requires PKCE for public clients and
+recommends it for confidential ones, and some providers reject an authorization
+code exchange that omits it. `golang.org/x/oauth2` implements it: generate a
+verifier per authorization request, send its challenge on the redirect, and hold
+the verifier until the callback.
 
 ```go
 func handleRedirect(w http.ResponseWriter, r *http.Request) {
-    http.Redirect(w, r, oauth2Config.AuthCodeURL(state), http.StatusFound)
+    // A fresh verifier is required for each authorization request. Store it the
+    // same way the state is stored, so the callback can replay it.
+    codeVerifier := oauth2.GenerateVerifier()
+
+    http.Redirect(w, r, oauth2Config.AuthCodeURL(state,
+        oauth2.S256ChallengeOption(codeVerifier)), http.StatusFound)
 }
 ```
+
+[pkce]: https://datatracker.ietf.org/doc/html/rfc7636
+[oauth2-bcp]: https://datatracker.ietf.org/doc/html/rfc9700#section-2.1.1
 
 Then, on the response, the ID Token verifier can be used to verify ID Tokens.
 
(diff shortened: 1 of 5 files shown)


ericchiang:
Thanks, but I don't have interest reviewing a PR with a Claude generated description. This doesn't inspire confidence that there's been thought put into it. And I can already generate PRs through Claude to review myself.

ryandiginomad:
Fair enough, and thanks for saying it directly.

The implementation and verification were mine: I ran all three examples against live discovery documents (Duende for the logout one, since Google publishes no `end_session_endpoint`) and checked the captured `code_challenge` against SHA256 of the verifier rather than just asserting the parameter was present.

Happy to reopen with a plain description if you'd look at the diff. If not, no hard feelings — the issue is still worth doing and I won't be offended if someone else picks it up.

PR #499, opened 2026-09-01
oidc: ignore JWKs with unsupported key types rather than failing

### LINKED ISSUE #498: RemoteKeySet fails on key sets containing Ed448 or X448 keys
An OpenID provider I'm integrating with serves a key set containing OKP keys with Ed448 and X448 curves and no "alg" member:

```json
{"kty":"OKP","use":"sig","crv":"Ed448","kid":"...","x":"..."}
```

go-jose returns `ErrUnsupportedKeyType` for these keys, `jwkJSON.UnmarshalJSON` propagates it, and `RemoteKeySet` rejects the whole set:

```
oidc: failed to decode keys: expected Content-Type = application/json, got "application/jwk-set+json; charset=UTF-8": go-jose/go-jose: unsupported key type/format ...
```

(The Content-Type in the message is a side effect of `unmarshalResp`, the unmarshal is what failed.)

The skip added in f9049c9 doesn't apply here because these keys carry no "alg". RFC 7517 section 5 says implementations should ignore keys they don't understand, and go-jose exports `jose.ErrUnsupportedKeyType` for that purpose. PR incoming.

nicoske:
Fixes #498

Skip keys whose unmarshal fails with `jose.ErrUnsupportedKeyType` instead of failing the whole set, same as the existing "alg" skip.

### DIFF
--- a/oidc/jwks.go
+++ b/oidc/jwks.go
@@ -278,6 +278,20 @@ func (j *jwkJSON) UnmarshalJSON(data []byte) error {
 		}
 		var jwk jose.JSONWebKey
 		if err := json.Unmarshal(key, &jwk); err != nil {
+			// Ignore keys with types that go-jose doesn't support, such as
+			// OKP keys with Ed448 or X448 curves. Some providers include
+			// them in their key sets without an "alg" value, so the check
+			// above doesn't catch them.
+			//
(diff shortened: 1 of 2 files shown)


ericchiang:
Thanks for the fix!

ericchiang:
This is now published https://github.com/coreos/go-oidc/releases/tag/v3.21.0

PR #500, opened 2026-09-06
oidc: reject ID tokens with a missing or empty sub claim

### LINKED ISSUE #495: IDTokenVerifier.Verify accepts ID Tokens with missing, null, or empty sub claim
## Summary

`IDTokenVerifier.Verify()` unmarshals the ID Token payload and returns success even when the `sub` (Subject) claim is missing, JSON `null`, or otherwise resolves to an empty string. OpenID Connect Core 1.0 defines `sub` as **REQUIRED** and the value as a **case-sensitive string**.

Callers that assume a successful `Verify()` implies a non-empty subject may therefore receive `Subject == ""` without an error. 

The issue was originally discussed with Red Hat, who also suggested that `UserInfo()` is also affected.

## Spec reference

OpenID Connect Core 1.0, [ID Token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken):

> **sub** REQUIRED. Subject Identifier. … The **sub value is a case-sensitive string**.

JSON `null` is not a string. An absent `sub` key leaves Go's `encoding/json` with the zero value for `string` (`""`). A conforming ID Token verification path should not treat either case as valid.

## Current behaviour

In `oidc/verify.go`, after signature verification, the payload is unmarshaled into an internal struct and `Subject` is copied onto the returned `IDToken` without a non-empty check:

```go
var token idToken
if err := json.Unmarshal(payload, &token); err != nil {
    return nil, fmt.Errorf("oidc: failed to unmarshal claims: %v", err)
}
// ... issuer, audience, expiry checks ...
t := &IDToken{
    Subject: token.Subject,
    // ...
}
return t, nil
```

**Cases that currently return `nil` error:**

| Payload | `IDToken.Subject` after `Verify` |
|---------|----------------------------------|
| `"sub": null` | `""` |
| `sub` key omitted | `""` |
| `"sub": ""` | `""` |

The library already rejects some malformed `sub` types (for example `"sub": ["a","b"]` fails unmarshaling into `string`).

The same `json.Unmarshal` pattern appears on the `UserInfo()` path; Red Hat noted that path may need the same kind of check.

## Precedent in the same package

`VerifyLogout()` in `oidc/logout.go` rejects tokens where both `Subject` and `SessionID` are empty. That suggests empty subject presence is worth validating; `Verify()` simply does not apply an equivalent guard today.

## Consistency with other checks in `Verify()`

`Verify()` already rejects tokens with wrong issuer, audience, or expiry. An empty or null `sub` is another OIDC ID Token requirement that could be checked at the same layer for consistency.

I understand the reluctance to add checks that break existing deployments. This change would only reject tokens that already violate the OIDC spec (no conforming OP should emit null or absent `sub`).

## Duplicate `sub` keys (related edge case)

If the payload contains duplicate JSON keys, for example `{"sub":"alice","sub":"bob",...}`, Go's `encoding/json` uses the **last** value when unmarshaling into a struct. Signature verification still passes. RFC 8259 recommends unique names but does not require them.

I am mentioning this for completeness. The null/absent `sub` case above is the simpler spec conformance fix. Duplicate-key handling may deserve separate discussion if you prefer a stricter JSON parse.

## Reproduction

Save the test from the **Appendix** below as `duplicate_sub_test.go` in the repo root, then run:

```
go test -v -run TestNullSub
go test -v -run TestDuplicateSub
```

`TestNullSub` signs a valid RS256 token with `"sub": null`, calls `Verify()`, and fails the test if `Subject == ""` with no error.

I can open a PR with this test and a minimal fix if that would help.

## Suggested fix (minimal)

After unmarshaling in `Verify()`, before returning:

```go
if token.Subject == "" {
    return nil, fmt.Errorf("oidc: token missing required 'sub' claim")
}
```

A similar non-empty `sub` check on `UserInfo()` would align both code paths. 

## Environment

Tested against `github.com/coreos/go-oidc/v3` v3.11.0. The same `json.Unmarshal` path is present on current v3 main.

## Appendix: self-contained test

```go
package oidc_test

// Self-contained reproducer for null/empty sub and duplicate sub keys.
// Save as duplicate_sub_test.go in the repo root and run:
//
//	go test -v -run TestNullSub
//	go test -v -run TestDuplicateSub

import (
	"context"
	"crypto/rand"
	"crypto/rsa"
	"encoding/json"
	"fmt"
	"net/http"
	"net/http/httptest"
	"testing"
	"time"

	"github.com/coreos/go-oidc/v3/oidc"
	"github.com/go-jose/go-jose/v4"
)

func setupTestOP(t *testing.T) (srv *httptest.Server, sign func(payload []byte) string) {
	t.Helper()

	key, err := rsa.GenerateKey(rand.Reader, 2048)
	if err != nil {
		t.Fatal(err)
	}
	const keyID = "k1"

	mux := http.NewServeMux()
	srv = httptest.NewServer(mux)
	t.Cleanup(srv.Close)

	mux.HandleFunc("/.well-known/openid-configuration", func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "application/json")
		fmt.Fprintf(w, `{"issuer":%q,"jwks_uri":%q}`, srv.URL, srv.URL+"/keys")
	})
	mux.HandleFunc("/keys", func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "application/json")
		jwks := jose.JSONWebKeySet{Keys: []jose.JSONWebKey{{
			Key:       &key.PublicKey,
			KeyID:     keyID,
			Algorithm: string(jose.RS256),
			Use:       "sig",
		}}}
		json.NewEncoder(w).Encode(jwks)
	})

	signer, err := jose.NewSigner(
		jose.SigningKey{Algorithm: jose.RS256, Key: key},
		(&jose.SignerOptions{}).WithHeader("kid", keyID),
	)
	if err != nil {
		t.Fatal(err)
	}

	sign = func(payload []byte) string {
		signed, err := signer.Sign(payload)
		if err != nil {
			t.Fatalf("sign: %v", err)
		}
		compact, err := signed.CompactSerialize()
		if err != nil {
			t.Fatalf("compact serialize: %v", err)
		}
		return compact
	}
	return
}

func newVerifier(t *testing.T, srv *httptest.Server) *oidc.IDTokenVerifier {
	t.Helper()
	ctx := oidc.InsecureAllowHTTP(context.Background())
	provider, err := oidc.NewProvider(ctx, srv.URL)
	if err != nil {
		t.Fatalf("NewProvider: %v", err)
	}
	return provider.Verifier(&oidc.Config{ClientID: "client"})
}

func TestNullSub(t *testing.T) {
	srv, sign := setupTestOP(t)

	now := time.Now()
	payload := fmt.Sprintf(
		`{"iss":%q,"sub":null,"aud":"client","exp":%d,"iat":%d}`,
		srv.URL, now.Add(time.Hour).Unix(), now.Unix(),
	)

	tok, err := newVerifier(t, srv).Verify(
		oidc.InsecureAllowHTTP(context.Background()), sign([]byte(payload)),
	)
	if err != nil {
		t.Fatalf("Verify returned error: %v", err)
	}
	if tok.Subject == "" {
		t.Errorf("Verify succeeded with empty Subject after sub:null payload")
	}
}

func TestDuplicateSub(t *testing.T) {
	srv, sign := setupTestOP(t)

	now := time.Now()
	// Two "sub" members. RFC 8259 recommends unique names; encoding/json uses the last value.
	payload := fmt.Sprintf(
		`{"iss":%q,"sub":"alice","sub":"bob","aud":"client","exp":%d,"iat":%d}`,
		srv.URL, now.Add(time.Hour).Unix(), now.Unix(),
	)

	tok, err := newVerifier(t, srv).Verify(
		oidc.InsecureAllowHTTP(context.Background()), sign([]byte(payload)),
	)
	if err != nil {
		t.Fatalf("Verify returned error: %v", err)
	}
	if tok.Subject != "alice" {
		t.Errorf("Subject=%q; expected first sub value or an error, got last unmarshaled value", tok.Subject)
	}
}
```

AshSgDe29071999:
OpenID Connect Core 1.0 requires `sub`. `Verify` returned success when the claim was absent, JSON `null`, or empty.

Fixes #495

## Test
`go test ./oidc -run TestVerify -count=1`

### DIFF
--- a/oidc/verify.go
+++ b/oidc/verify.go
@@ -220,6 +220,10 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string) (*IDTok
 		distributedClaims[cn] = s
 	}
 
+	if token.Subject == "" {
+		return nil, fmt.Errorf("oidc: id token missing required 'sub' claim")
+	}
+
 	t := &IDToken{
 		Issuer:            token.Issuer,
 		Subject:           token.Subject,
--- a/oidc/verify_test.go
+++ b/oidc/verify_test.go
@@ -16,8 +16,28 @@ import (
 func TestVerify(t *testing.T) {
 	tests := []verificationTest{
 		{
-			name:    "good token",
+			name:    "missing sub",
 			idToken: `{"iss":"https://foo"}`,
+			config: Config{
+				SkipClientIDCheck: true,
+				SkipExpiryCheck:   true,
+			},
+			signKey: newRSAKey(t),
+			wantErr: true,
+		},
+		{
+			name:    "null sub",
+			idToken: `{"iss":"https://foo","sub":null}`,
+			config: Config{
+				SkipClientIDCheck: true,
+				SkipExpiryCheck:   true,
+			},
+			signKey: newRSAKey(t),
+			wantErr: true,
+		},
+		{
+			name:    "good token",
+			idToken: `{"iss":"https://foo","sub":"user"}`,
 			config: Config{
 				SkipClientIDCheck: true,
 				SkipExpiryCheck:   true,
@@ -26,7 +46,7 @@ func TestVerify(t *testing.T) {
 		},
 		{
 			name:    "good eddsa token",
-			idToken: `{"iss":"https://foo"}`,
+			idToken: `{"iss":"https://foo","sub":"user"}`,
 			config: Config{
 				SkipClientIDCheck:    true,
 				SkipExpiryCheck:      true,
@@ -37,7 +57,7 @@ func TestVerify(t *testing.T) {
 		{
 			name:    "invalid issuer",
 			issuer:  "https://bar",
-			idToken: `{"iss":"https://foo"}`,
+			idToken: `{"iss":"https://foo","sub":"user"}`,
 			config: Config{
 				SkipClientIDCheck: true,
 				SkipExpiryCheck:   true,
@@ -48,7 +68,7 @@ func TestVerify(t *testing.T) {
 		{
 			name:    "skip issuer check",
 			issuer:  "https://bar",
-			idToken: `{"iss":"https://foo"}`,
+			idToken: `{"iss":"https://foo","sub":"user"}`,
 			config: Config{
 				SkipIssuerCheck:   true,
 				SkipClientIDCheck: true,
@@ -58,7 +78,7 @@ func TestVerify(t *testing.T) {
 		},
 		{
 			name:    "invalid sig",
-			idToken: `{"iss":"https://foo"}`,
+			idToken: `{"iss":"https://foo","sub":"user"}`,
 			config: Config{
 				SkipClientIDCheck: true,
 				SkipExpiryCheck:   true,
@@ -70,7 +90,7 @@ func TestVerify(t *testing.T) {
 		{
 			name:    "google accounts without scheme",
 			issuer:  "https://accounts.google.com",
-			idToken: `{"iss":"accounts.google.com"}`,
+			idToken: `{"iss":"accounts.google.com","sub":"user"}`,
 			config: Config{
 				SkipClientIDCheck: true,
 				SkipExpiryCheck:   true,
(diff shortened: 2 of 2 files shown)


ericchiang:
https://github.com/coreos/go-oidc/issues/495#issuecomment-4948756145

